PR #7, opened 2021-12-24
state: replace serde_json with bincode to support deserialization of non-string keys

yaa110:
when using serde_json, after creation of tokens and pairs, deserialization is failed by the error of `key must be a string` because we are using custom keys for canisters hashmap. This PR fixes that issue by serializing using `bincode`.


Maximkaaa:
I don't think that using bincode is the best solution for serializing state, for it doesn't contain schema information, so if the state structs change at some point, it will be very difficult to catch bugs and debug this.

Stable storage actually uses candid to serialize and deserialize data, so it's better to use candid for that. This can be done by deriving `CandidType` and `Deserialize` on the `State` struct, and then you can store and restore it by simply

```
let state = State::new();
ic_cdk::storage::stable_save(&state).unwrap();

let state = ic_cdk::storage::stable_restore().unwrap();
```

yaa110:
@Maximkaaa using `CandidType` also has the same issue and if a new field is added at some point, It could not be deserialized.
Moreover, `CandidType` is not implemented for many other types such as `PhantomData`, so, we need to implement extensions for those types that makes it difficult to use `CandidType`.

Maximkaaa:
Candid serialization includes header, which allows it to explain you in the error, what exactly went wrong during deserialization. And in most cases changing u32 to u64 will not break canid, but will break bincode.

About the `PhantomData` issue, I agree about that. But I don't think PhantomData is needed for factory state. I'll write a suggestion for some refactoring/improvement of these helpers on Monday.

yaa110:
How can `CandidType` help with addition or deletion of new fields? I think a change from `u32` to `u64` is rare, however, adding or deleting a field might be done frequently. It seems changing from `bincode` to `CandidType` has no benefit and adds more complexity and restrictions. Maybe, we need to find a better way to serialize that is not schemaless and also has no restriction in key type.

yaa110:
Maybe it is better to have string keys and use json serde

PR #15, opened 2022-01-11
[CPROD-379] Adds pair principal extension

Maximkaaa:


canisters/src/pair/principal_ext.rs
```diff
@@ -0,0 +1,18 @@
+use ic_cdk::export::candid::Principal;
+use async_trait::async_trait;
+
+#[async_trait]
+pub trait PairPrincipalExt {
+    /// Sets the liquidity cap for the pair. This method can only be called by the pair owner,
+    /// which is usually the pair factory.
+    async fn set_cap(&self, amount: Option<u128>) -> Result<(), String>;
```
blutooth, line 8: Just one question - is a get cap needed?

Maximkaaa: Get cap is not used by factory, so no need for it in helpers. Pair canister has this method.

PR #16, opened 2022-01-13
[CPROD-476] Add common types for AMM pair

Maximkaaa:


canisters/src/pair/types.rs
```diff
@@ -0,0 +1,39 @@
...
+
+impl Standard {
+    pub fn as_str(&self) -> &str {
+        match self {
+            Self::Ledger => "ledger",
+            Self::Erc20 => "erc20",
+        }
+    }
+
+    pub fn from_str(s: &str) -> Self {
```
hagsteel, line 17: The pedantic in me would say implement this instead of having a from_str function, especially since this can fail: https://doc.rust-lang.org/std/str/trait.FromStr.html
However this is just a suggestion

hagsteel (approved):

canisters/src/pair/types.rs
Maximkaaa: Yeah, I though about using ToStr and FromStr traits, but it would add unnecessary complications. I hope to get rid of this altogether in future anyway.

PR #24, opened 2022-02-07
[CPROD-542] Adds methods for using cycles for canister creation by factory

Maximkaaa:
The logic of the `FactoryState::get_provided_cycles` is as follows:
1. Check if the current canister call was done with cycles (through a wallet). If it is so, the factory uses the provided cycles to create a canister, using all the cycles given minus some fixed fee to cover the factory expenses.
2. Otherwise, the factory checks the balance of the caller subaccount in the ICP ledger. If there is enough ICP there (meaning, that the caller transferred the needed ICP beforehand), the canister is created with ICP:
  2.1. Set amount of ICP (`icp_fee`) is transferred to the `icp_to` principal.
  2.2. The canister is created using `MIN_CANISTER_CYCLES` cycles from the factory balance.
3. Otherwise, an error is return asking for $$$.


ic-helpers/src/ledger/principal_ext.rs
```diff
@@ -37,90 +29,119 @@ pub struct Block{
...
+            return Err(format!("cannot transfer tokens: amount '{}' is smaller then the fee '{}'", amount, TRANSACTION_FEE.get_e8s()))
+        }
+
+        let args = TransferArgs {
+            memo: Default::default(),
+            amount: (Tokens::from_e8s(amount) - TRANSACTION_FEE)?,
+            fee: TRANSACTION_FEE,
+            from_subaccount,
+            to: AccountIdentifier::from_principal(to, to_subaccount).to_address(),
+            created_at_time: None
```
hagsteel, line 138: Should this be set?

ic-helpers/src/factory/state.rs
```diff
@@ -40,3 +51,93 @@ macro_rules! impl_factory_state_management {
         }
     };
 }
+
+/// This trait must be implemented by a factory state to make using of `init_factory_api` macro
+/// possible.
+pub trait FactoryState<K: Hash + Eq> {
```
hagsteel, line 57: As per our conversation: add an inner concrete type for the three fields

hagsteel (commented):
Rust fmt should get some love here ;)

ic-helpers/src/ledger/principal_ext.rs
Maximkaaa: Doesn't need to be set. IC sets it to the current TS for the caller anyway.

AlessandroRiet (approved):

PR #33, opened 2022-03-24
Cprod 682

hagsteel:
* Adding versioned storage to the canister derive macro

### DIFF
--- a/ic-canister/ic-canister-macros/src/derive.rs
+++ b/ic-canister/ic-canister-macros/src/derive.rs
@@ -3,8 +3,8 @@ use quote::quote;
 use std::collections::HashSet;
 use syn::parse::{Parse, ParseStream};
 use syn::{
-    parse_macro_input, Attribute, Data, DeriveInput, Field, Fields, GenericArgument, Path,
-    PathArguments, Type,
+    parse_macro_input, Attribute, Data, DeriveInput, Field, Fields, GenericArgument, Lit, LitBool,
+    Meta, NestedMeta, Path, PathArguments, Type,
 };
 
 #[derive(Debug)]
@@ -22,8 +22,8 @@ impl Parse for TraitNameAttr {
 impl Default for TraitNameAttr {
     fn default() -> Self {
         let tokens = TokenStream::from(quote! {::ic_canister::Canister});
-        let path = parse_macro_input::parse::<Path>(tokens)
-            .expect("static value parsing always succeeds");
+        let path =
+            parse_macro_input::parse::<Path>(tokens).expect("static value parsing always succeeds");
         Self { path }
     }
 }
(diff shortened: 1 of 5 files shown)


ic-canister/ic-canister-macros/src/derive.rs
```diff
@@ -213,28 +201,38 @@ fn expand_upgrade_methods(
...
+                ::ic_storage::stable::write(#state_borrow).unwrap();
             }
 
             #[cfg(all(target_arch = "wasm32", feature = "export_api"))]
             #[export_name = "canister_post_upgrade"]
             fn __post_upgrade() {
                 use ::ic_storage::IcStorage;
 
-                let (#( #field_names,)*) = ::ic_cdk::storage::stable_restore().unwrap();
+                let #name = match ::ic_storage::stable::read::<#field_type::Versioned>() {
```
Maximkaaa, line 214: Shouldn't this be `::Previous`?

ic-canister/ic-canister-macros/src/derive.rs
```diff
@@ -213,28 +201,38 @@ fn expand_upgrade_methods(
...
+                    Ok(val) => val,
+                    Err(e) => panic!("failed to upgrade: {}", e),
```
Maximkaaa, line 216: For debugging purposes, I think it's better to use `::ic_cdk::trap` here

ic-canister/ic-canister-macros/src/derive.rs
```diff
@@ -213,28 +201,38 @@ fn expand_upgrade_methods(
...
 
-                #( #fields_assignment )*
+                #field_assignment
             }
         }
     }
 }
 
 fn is_state_field_stable(_field: &Field) -> bool {
+    // path = stable_store
```
Maximkaaa, line 226: Do we need this commented stuf?

ic-canister/ic-canister-macros/src/derive.rs
```diff
@@ -103,8 +103,8 @@ pub fn derive_canister(input: TokenStream) -> TokenStream {
                 .push(quote! {#field_name : <#field_type as ::ic_storage::IcStorage>::get()});
             state_fields_test.push(quote! {#field_name : ::std::rc::Rc::new(::std::cell::RefCell::new(<#field_type as ::std::default::Default>::default()))});
 
-            if is_stable {
-                stable_fields.push((field_name, field_type));
+            if is_stable && stable_field.is_none() {
```
Maximkaaa, line 106: Doesn't ths mean that if the `State` struct has more then one stable field, the macro will just disregard all except the first one. I think we want to throw a compile error in this case.

ic-canister/ic-canister-macros/src/derive.rs
hagsteel: Good spot!
Changed and updated

ic-canister/ic-canister-macros/src/derive.rs
hagsteel: Good point

ic-canister/ic-canister-macros/src/derive.rs
hagsteel: Fixed

ic-canister/ic-canister-macros/src/derive.rs
hagsteel: Nope, implemented the function instead.

blutooth (approved):

blutooth (approved):

PR #34, opened 2022-03-30
Fix upgrades

hagsteel:
* Change the canister upgrade functions to use an instance.
* Fix `trap` (had incorrect args)
* Check architecture rather than test for local testing functions

### DIFF
--- a/ic-canister/ic-canister-macros/src/derive.rs
+++ b/ic-canister/ic-canister-macros/src/derive.rs
@@ -197,7 +197,7 @@ fn expand_upgrade_methods(
         quote! { &* #name.borrow(), },
     );
 
-    let field_assignment = quote! { #field_type::get().replace(#name); };
+    let field_assignment = quote! { self. #name.replace(#name); };
 
     quote! {
         impl #struct_name {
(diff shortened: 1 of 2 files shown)


ic-canister/ic-canister-macros/src/derive.rs
```diff
@@ -193,36 +193,48 @@ fn expand_upgrade_methods(
...
         impl #struct_name {
-            #[cfg(all(target_arch = "wasm32", feature = "export_api"))]
+            #[cfg(feature = "export_api")]
             #[export_name = "canister_pre_upgrade"]
             fn __pre_upgrade() {
+                let instance = Self::init_instance();
+                instance.__pre_upgrade_inst();
+            }
+
+            fn __pre_upgrade_inst(&self) {
```
Maximkaaa, line 211: One thing we could consider doing is to add these methods to the `Canister` trait, so that we don't have to have magic `__pre_upgrade_inst` in the tests.

In case there's no stable state, we can just generate empty methods (as oposed to skipping them altogether in current implementation)

Maximkaaa (approved):

AlessandroRiet (approved):

PR #35, opened 2022-03-31
[WPROD-53] Add `extend-with-factory-api` macro

awkure:
Problem: When building canisters from `ic-canister::Canister` trait
we have no api to generate factory endpoints to it.

Solution: Add `extend-with-factory-api` macro that takes the
existing canister and extends it with the api needed.

### DIFF
--- a/ic-helpers/src/factory/api.rs
+++ b/ic-helpers/src/factory/api.rs
@@ -25,6 +25,10 @@
 /// * set_controller
 /// * refund_icp
 #[macro_export]
+#[deprecated(
+    since = "0.2.1",
+    note = "use `extend_with_factory_api` that extends `ic-canister::Canister` trait instead"
+)]
 macro_rules! init_factory_api {
     ( $state:ident, $bytecode:expr ) => {
         /// Returns the checksum of a wasm module in hex representation.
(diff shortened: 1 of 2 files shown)


ic-helpers/src/factory/api.rs
```diff
@@ -210,3 +214,217 @@ macro_rules! init_factory_api {
...
+/// * get_controller
+/// * set_controller
+/// * refund_icp
+#[macro_export]
+macro_rules! extend_with_factory_api {
+    ( $canister:ident, $upgrading_bytecode:expr ) => {
+        impl $canister {
+            /// Returns the checksum of a wasm module in hex representation.
+            #[query]
+            #[candid_method(query)]
```
Maximkaaa, line 242: `candid_method` macros are not needed after this change, so we should remove it.

ic-helpers/src/factory/api.rs
```diff
@@ -210,3 +214,217 @@ macro_rules! init_factory_api {
...
+            fn get_checksum(&self) -> String {
+                self.state.borrow().factory().checksum.to_string()
```
Maximkaaa, line 244: The state parameter is not necesserilly called `state` in the canister. It can be any other name. So it's better to have this name as the parameter of the macro, and write `self. $state` here intead.

Maximkaaa (changes requested):

Maximkaaa (approved):

hagsteel (approved):

PR #36, opened 2022-04-04
[WPROD-53] Add random principal ids for the Canister generated structs

awkure:
This had names clashing if you create token canisters like in the example below
```rust
mod Token0 {
    #[derive(Clone, Canister)]
    pub struct Token0Canister {
        #[id]
        principal: Principal,
    }
}

mod Token1 {
    #[derive(Clone, Canister)]
    pub struct Token1Canister {
        #[id]
        principal: Principal,
    }
}
```


awkure:
The only problem is that it forces the user of the canister to add `rand` dependency.

PR #39, opened 2022-04-05
[CPROD-724] Add mocking calls for the management canister

awkure:
Since we want to be able to use mocking api for pair factories in testing, we need to make management canister methods being mocked as well. This is allowed from `virtual_canister_call!` macro that generates the necessary code depending on the target arch being compiled. This PR addresses the issue.


ic-helpers/src/macros.rs
```diff
@@ -0,0 +1,49 @@
+#[macro_export]
```
Maximkaaa, line 1: Instead of all this it's better to just use `ic-kit` crate. It does the same thing, but better.

ic-helpers/src/factory/state.rs
```diff
@@ -160,7 +160,19 @@ async fn consume_provided_cycles_or_icp(
     icp_fee: u64,
     controller: Principal,
 ) -> Result<u64, FactoryError> {
-    if ic_cdk::api::call::msg_cycles_available() > 0 {
+    if {
```
Maximkaaa, line 163: This also can be rewritten with `ic-kit`

awkure: So, we're not the first to stumble upon this problem :p

hagsteel (approved):
Looks good

Maximkaaa (approved):

PR #41, opened 2022-04-11
Replase some `ic_cdk` calls with `ic_kit` alternatives to allow mocking

awkure:


hagsteel (approved):

ic-helpers/src/ledger/principal_ext.rs
```diff
@@ -154,20 +154,28 @@ impl LedgerPrincipalExt for Principal {
...
+            let result = ic_cdk::call::<_, (Tokens,)>(*self, "account_balance", (args,))
+                .await
+                .map_err(|e| e.1)?
+                .0;
+            Ok(result.get_e8s())
+        }
+        #[cfg(not(target_family = "wasm"))]
+        {
+            Ok(10u64.pow(15))
+        }
```
Maximkaaa, line 178: This behavior would be very unexpected for anyone using this code. It also make it impossible to test cases when the balance is not sufficient etc.

When the factory creates a canister, there's logic there not to check the balance or consume ICP in case the caller of the request is the factory controller. If you use that logic for your tests, this hack will not be needed.

To mock the calller principal, you can use ic-kit library.

ic-helpers/src/factory/state.rs
```diff
@@ -206,21 +206,25 @@ async fn consume_provided_icp(
...
+        LedgerPrincipalExt::transfer(
+            &ledger,
+            icp_to,
+            amount,
+            Some(Subaccount::from(&PrincipalId::from(from))),
+            None,
+        )
+        .await
+        .map_err(FactoryError::LedgerError)?;
+    }
```
Maximkaaa, line 228: Same as above comment

ic-helpers/src/factory/state.rs
```diff
@@ -23,11 +23,11 @@ macro_rules! impl_factory_state_management {
     ( $state:ident, $bytecode:expr ) => {
         impl $state {
             pub fn stable_save(&self) {
-                ::ic_cdk::storage::stable_save((self,)).unwrap();
+                ::ic_kit::ic::stable_store((self,)).unwrap();
```
hagsteel, line 26: Can we get rid of these unwraps. Either use `expect` if it can't fail, otherwise we should return a `Result` here

ic-helpers/src/factory/state.rs
```diff
@@ -23,11 +23,11 @@ macro_rules! impl_factory_state_management {
...
             }
 
             pub fn stable_restore() {
-                let (mut loaded,): (Self,) = ::ic_cdk::storage::stable_restore().unwrap();
+                let (mut loaded,): (Self,) = ::ic_kit::ic::stable_restore().unwrap();
```
hagsteel, line 30: Can we get rid of these unwraps. Either use expect if it can't fail, otherwise we should return a Result here

hagsteel (changes requested):

ic-helpers/src/ledger/principal_ext.rs
awkure: Yeah, I didn't bother with it much because the ledger canister will be rewritten as a canister anyway. It was problematic to rewrite it into `ic-kit` call because of 

```
future cannot be sent between threads safely
the trait `std::marker::Send` is not implemented for `dyn Future<Output = Result<(ledger_canister::Tokens,), (ic_kit::RejectionCode, std::string::String)>>`
required for the cast to the object type `dyn Future<Output = Result<u64, std::string::String>> + std::marker::Send`rustc
[principal_ext.rs(171, 22): ]()future is not `Send` as it awaits another future which is not `Send`
```

ic-helpers/src/factory/state.rs
awkure: Again, it'll be more clear when the ledger will be rewritten to have a more elaborate api

ic-helpers/src/factory/state.rs
awkure: Sure we can. My reasoning was that if they're present, then this was intentional. From my perspective, ic canisters are prone to use unwraps instead of a proper error handling because I've been seeing this everywhere with the caption that it allows to kill the thread and revert the canister state.

awkure: Also imo this macro is unneeded in first place because this separation of the state api between the crates is more confusing and probably unnecessary due to the way `ic-canister` canisters handle state

ic-helpers/src/ledger/principal_ext.rs
Maximkaaa: It's still ok, for you don't need to use ic_kit in here. Just leave ic_cdk api call as is. When testing, this code will never be executed because of the logic I described above.

awkure: Yep, seem to be working. Thanks. To avoid further confusion we need to document it somewhere though.

ic-helpers/src/factory/state.rs
blutooth: probably is no longer needed..

ic-helpers/src/factory/state.rs
blutooth: probably we can delete he macro?

hagsteel (approved):

blutooth (approved):

PR #44, opened 2022-04-27
[CPROD-753] Update ic dependency, refactoring

Maximkaaa:
Update ic fork dependency to a more recent branch, to make it possible to use with IC. This also required a little refactoring (renaming `TRANSACTION_FEE` to `DEFAULT_TRANSFER_FEE`.

Also added better error messages for derive macro and added possibility to retrieve canister principal id from inside the canister with simple `self.principal()` call.

### DIFF
--- a/ic-helpers/src/ledger/principal_ext.rs
+++ b/ic-helpers/src/ledger/principal_ext.rs
@@ -1,14 +1,14 @@
 use crate::ledger::account_id::FromPrincipal;
 use async_trait::async_trait;
-use dfn_core::api::call_with_cleanup;
+use dfn_core::api::{call_with_cleanup, PrincipalId};
+use dfn_core::CanisterId;
 use dfn_protobuf::protobuf;
 use ic_cdk::export::candid::{CandidType, Principal};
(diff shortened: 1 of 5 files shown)


ic-helpers/src/ledger/principal_ext.rs
```diff
@@ -177,18 +177,18 @@ impl LedgerPrincipalExt for Principal {
         from_subaccount: Option<Subaccount>,
         to_subaccount: Option<Subaccount>,
     ) -> Result<u64, String> {
-        if amount < TRANSACTION_FEE.get_e8s() {
+        if amount < DEFAULT_TRANSFER_FEE.get_e8s() {
             return Err(format!(
                 "cannot transfer tokens: amount '{}' is smaller then the fee '{}'",
```
hagsteel, line 182:
```suggestion
                "cannot transfer tokens: amount '{}' is less then the fee '{}'",
```

hagsteel (approved):
Wording change since this is a public canister, but it's not a show stopper

blutooth (approved):

PR #45, opened 2022-04-28
[CPROD-736] Signing calls to management canister

awkure:


ic-canister/ic-canister/src/lib.rs
```diff
@@ -452,8 +452,14 @@ pub fn call_virtual_responder(
...
-                    method_name, principal
+                    "canister method {} is not registered for principal {}, METHODS: {:?}",
+                    method_name,
+                    principal,
+                    responders
+                        .borrow()
+                        .keys()
+                        .map(|(k, r)| (k.to_string(), r))
+                        .collect::<Vec<_>>()
                 ),
```
hagsteel, line 463: You can move the `method_name` and `principal` directly into the string it self:

```rust
                format!(
                    "canister method {method_name} is not registered for principal {principal}, METHODS: {:?}",
                    responders
                        .borrow()
                        .keys()
                        .map(|(k, r)| (k.to_string(), r))
                        .collect::<Vec<_>>()
                ),
```

ic-helpers/src/management/types.rs
```diff
@@ -0,0 +1,57 @@
...
+/// This is a structure that is only used for `get_ecdsa_pubkey` call in
+/// management canister. The only problem that it is only used by dfx in
+/// November build of ic, but as a dependency we use the more recent one,
+/// hence, to be able to deserialized the result of `get_ecdsa_pubkey`
+/// we need something that represents its response.
+#[derive(CandidType, Deserialize, Debug)]
+pub struct GetECDSAPublicKeyResponse {
+    pub public_key: Vec<u8>,
+    pub chain_code: Vec<u8>,
+}
```
hagsteel, line 43: Can this be moved to management/canister.rs and be made private instead, or is another project depending on this as well from this crate?

hagsteel (changes requested):

ic-helpers/src/management/types.rs
awkure: What do you mean exactly? Are there any fields that we want to keep private? This struct is used for deserialisation only and I used it token for extracting fields from but of course we could return something else.

hagsteel: What I mean is: if this struct is only used in one place, we don't need it in types.rs (which is something I hope we can get rid of at some point) and it cane be moved to the module that is using it. Also no need to make any of it public for the same reason.

awkure: This struct is exported and used in outside calls for deserialization

ic-helpers/src/management/canister.rs
```diff
@@ -143,6 +145,41 @@ impl Canister {
         .map(|r| Self(r.canister_id))
     }
 
+    pub async fn get_ecdsa_pubkey(
+        principal: Option<Principal>,
+    ) -> Result<Pubkey, (RejectionCode, String)> {
+        let request = GetECDSAPublicKeyArgs {
+            canister_id: principal,
+            derivation_path: vec![],
```
Maximkaaa, line 153: We will need to use this parameter in the `neuro` project, so let's have it as an argument to this method. It's not a big deal to give empty vector as an argument when it's not needed.

ic-helpers/src/management/canister.rs
```diff
@@ -143,6 +145,41 @@ impl Canister {
...
+        .map(|bytes| Pubkey::new(bytes.public_key))
+    }
+
+    pub async fn sign_with_ecdsa(
+        hash: Vec<u8>,
+    ) -> Result<SignWithECDSAReply, (RejectionCode, String)> {
+        let request = SignWithECDSAArgs {
+            key_id: "secp256k1".into(),
+            message_hash: hash,
+            derivation_path: vec![],
```
Maximkaaa, line 172: This also needs to be an argument to the method.

ic-helpers/src/management/types.rs
```diff
@@ -0,0 +1,57 @@
...
+        Self(Vec::new())
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+#[derive(CandidType, Deserialize, Debug)]
+pub struct GetECDSAPublicKeyArgs {
```
Maximkaaa, line 28: Can't we use this structure from the `ic` dependency? Why do we duplicate it here?

Maximkaaa: The same applies to other structures below

ic-helpers/src/management/types.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub key_id: String,
+}
+
+/// This is a structure that is only used for `get_ecdsa_pubkey` call in
+/// management canister. The only problem that it is only used by dfx in
+/// November build of ic, but as a dependency we use the more recent one,
+/// hence, to be able to deserialized the result of `get_ecdsa_pubkey`
+/// we need something that represents its response.
+#[derive(CandidType, Deserialize, Debug)]
+pub struct GetECDSAPublicKeyResponse {
```
Maximkaaa, line 40: I believe this structure was renamed to `ECDSAPublicKeyResponse`, so we can use the original version from `ic`. It is present in our fork of the repo.

Maximkaaa (changes requested):

ic-helpers/src/management/types.rs
awkure: They're removed with the current version of ic that we use

awkure:
I personally don't like using `rebased-2022-02-17` branch because of these commits
image
But when I tried do a clean update of the ic fork, for some reason `ledger_canister` could not be found in remote ([the discussion](https://infinityswap.slack.com/archives/C02P69J279T/p1651579493630239)).

Maximkaaa (approved):

hagsteel (approved):

PR #54, opened 2022-05-23
Oneshot

aewc:
add the Oneshot(aka Oneway) macro


blutooth (commented):
Hi Liquan, can we add a couple of small unit tests for this. Thank you.

ic-canister/ic-canister-macros/src/lib.rs
```diff
@@ -19,6 +19,11 @@ pub fn canister_call(input: TokenStream) -> TokenStream {
     canister_call::canister_call(input)
 }
 
+#[proc_macro]
+pub fn canister_call_oneway(input: TokenStream) -> TokenStream {
```
Maximkaaa, line 23: I suggest we call the macro `canister_notify`.

ic-canister/tests/canister_b/src/lib.rs
```diff
@@ -55,6 +55,29 @@ impl CanisterB {
...
+    #[allow(unused_mut)]
+    #[allow(unused_variables)]
+    async fn call_increment_oneway(&self, value: u32) -> bool {
+        let mut canister_a = CanisterA::from_principal(self.state.borrow().canister_a);
+
+        #[cfg(target_arch = "wasm32")]
+        canister_call_oneway!(canister_a.inc_counter(value), ()).unwrap();
+
+        #[cfg(not(target_arch = "wasm32"))]
+        canister_call_oneway!(canister_a.inc_counter(value), ()).await.unwrap();
```
Maximkaaa, line 69: The usage of the macro must be same in both environments. So both these calls must be not `async`.

ic-canister/ic-canister-macros/src/lib.rs
Maximkaaa: Also, please, add documentation for new public macros in the crate. And also, to the crate level documentation of `ic-canister` crate.

Maximkaaa (changes requested):

ic-canister/ic-canister-macros/src/lib.rs
aewc: Using canister_notify is better for consistency with ic-cdk.
Before I thought its essence is oneway call, ic-cdk named notify is not a good idea. Anyway, let's use canister_notify.

Thanks for reminder about the document, I will refer to other interfaces to improve.

ic-canister/tests/canister_b/src/lib.rs
aewc: Yes, I'll figure out the mock calls in a non-wasm environment, and change canister_notify to be non-async

ic-canister/ic-canister-macros/src/api.rs
```diff
@@ -138,6 +140,14 @@ pub(crate) fn api_method(
             // todo: trap handler
             Ok(self. #method(#args_destr) #await_call)
         }
+
+        #[cfg(not(target_arch = "wasm32"))]
+        #[allow(unused_mut)]
+        pub fn #internal_method_notify(#args) -> Result<(), ::ic_cdk::api::call::RejectionCode> {
```
aewc, line 146: There are lots of warnings building the IS20 after I add this method:
```bash
warning: unused `Result` that must be used
   --> src/token/src/canister.rs:246:5
    |
246 |     #[update]
    |     ^^^^^^^^^
    |
    = note: this `Result` may be an `Err` variant, which should be handled
    = note: this warning originates in the attribute macro `update` (in Nightly builds, run with -Z macro-backtrace for more info)
```

PR #57, opened 2022-05-30
Fix dependencies to ic repo

Maximkaaa:


Maximkaaa:
I pushed another commit to this PR fixing a few more issues:

* Fix feature usage on upgrade methods in api derive macros. Because of wrong `cfg` statement our canisters' uprade was totally broken (all the state was lost)
* Move some of the pair types into the pair module (out of types.rs) to fix build of the AMM pair. It was broken by some long-ago refactoring.
* Correct the return type of notify method in api derive macro to fix pair factory build. It has different `Result` type in scope, so return type without specifying namespace broke the build.

AlessandroRiet (approved):

PR #56, opened 2022-05-27
Collect metrics for the `ic-canister`-derived canisters

awkure:

PR #58, opened 2022-05-31
[CPROD-846] Collect metrics for the ic canisters

awkure:
This is the same as https://github.com/infinity-swap/ic-helpers/pull/56 but without using deriving macros


Maximkaaa (changes requested):
It seems to me that in this form the `Metrics` are not really related to `ic-canister`, so it's a little strange to have documentation for them in that crate. Why don't we move the docs to the `MetricMap` struct.

It's probably fine to test this in `CanisterA`/`CanisterB`, though the test you wrote should rather be just a simple unit test for that struct. That would be cleaner. But if you leave it as is, I won't mind too much.

Maximkaaa:
I don't think it's a good approach for docs. We don't put factory docs, ledger and management API docs in here because there is more appropriate places for them. Same goes with metrics.

Maximkaaa (approved):

AlessandroRiet (approved):

PR #61, opened 2022-06-02
Remove `extend_with_factory_api!()` macro and set up tagless final api for treating traits as canisters

awkure:
This PR adds macros to treat trait as canisters to be able to inherit canister methods as trait implementations, e.g. `TokenFactoryCanister: FactoryCanister`.


ic-canister/ic-canister-macros/src/api.rs
```diff
@@ -346,7 +469,13 @@ fn get_args(sig: &Signature) -> Result<(Vec<Type>, Vec<Type>), Error> {
         ReturnType::Default => Vec::new(),
         ReturnType::Type(_, ty) => match ty.as_ref() {
             Type::Tuple(tuple) => tuple.elems.iter().cloned().collect(),
-            _ => vec![ty.as_ref().clone()],
+            ty => {
+                // Some types in trait canisters had to be marked as `AsyncReturn` as implementation detail
+                // but we do not need this when exporting them to candid files as ic calls them correctly
+                // in any case.
+                let extracted_type = crate::derive::extract_type_if_matches("AsyncReturn", ty);
+                vec![extracted_type.clone()]
```
Maximkaaa, line 477: Can we support here both using the alias and the original type (`Box<Pin<...>>`)?

Maximkaaa (changes requested):
The code looks good. I think this change requires very thorough documenting, otherwise we will not remember how to use it in couple of months.

Please, add some comments in the `ic-canister` top module how to use canister trait composition, mention idl merging and generate_exports! macro. A short explanation about why it was implemented this way would also be useful. An example of using the traits with idl merging and inter-crate trait inheritance could also be useful.

ic-canister/ic-canister-macros/src/api.rs
awkure: Sure. Regarding docs it was planned, yes, after I deal with integrational tests in amm

ic-factory/src/api.rs
```diff
@@ -229,13 +229,17 @@ impl FactoryCanister for FactoryExport {
...
-    fn state_mut(&self) -> RefMut<FactoryState> {
-        self.state.borrow_mut()
-    }
+    //     fn state(&self) -> Ref<FactoryState> {
+    //         self.state.borrow()
+    //     }
+
+    //     fn state_mut(&self) -> RefMut<FactoryState> {
+    //         self.state.borrow_mut()
+    //     }
```
Maximkaaa, line 242: вуд

Maximkaaa (approved):
Very nice documentation. Good job.

blutooth (approved):

PR #67, opened 2022-06-21
Changes to upgrade methods

awkure:


AlessandroRiet (changes requested):
As discussed, we've got avoidable code duplication here.
PR description also needs to be written.

awkure:
We had a problem of duplicated upgrade calls for the canister when generated a candid file for the amm canister. This was partly solved by https://github.com/infinity-swap/canister-sdk/pull/65, but we also need those upgrade calls when testing canisters via `ic-canister` framework, hence code duplication because `#[export_name]` macro depends on the target architecture of the call

PR #68, opened 2022-06-22
[CPROD-899] Add attribute to allow custom upgrade methods in ic-canister

Maximkaaa:
This PR changes the way the canister state is instantiated in test context. Instead of keeping the state for each canister instance in a structure generated by `Canister` derive macro, we move this HashMap to the implementation of ic-storage macros.

This way the state can safely be used as is in traits and structs alike. E.g. every call to `IcStorage::get()` will always consider test context and produce correct result both in test environment and in wasm canister.

Using this approach, we can now easily store the factory state inside the trait, exposing it by the factory trait method.

With this approach implementing custom upgrade methods for canisters using multiple state structures becomes much easier. To allow this we introduce `#[canister_no_upgrade_methods]` attribute that instructs the derive macro not to generate the upgrade methods.


ic-canister/tests/canister_a/src/lib.rs
```diff
@@ -6,21 +6,24 @@ use std::{cell::RefCell, rc::Rc};
...
 }
 
-pub trait CanisterA {
-    fn state(&self) -> Rc<RefCell<State>>;
+pub trait CanisterA: ic_canister::Canister {
+    fn state(&self) -> Rc<RefCell<StateA>> {
+        StateA::get()
+    }
 
+    /// Documentation for get_counter
```
awkure, line 26: change?

awkure:
So, how this particularly solves the issue with trait factory state upgrading in case of is20? We mark is20 factory with `#[canister_no_upgrade_methods]` and manually set up upgrade methods for trait factory state? How is trait factory state gonna be preserved in is20 factory?

PR #69, opened 2022-06-22
[CPROD-922] Pre-update hook for `ic-canister`-derived canisters

awkure:
We want to be able to run some validation or collect metrics before executing *any* update call. This PR adds `pre_update` hook that can be overwritten by the user if he decides to do so.


ic-canister/tests/canister_a/src/lib.rs
```diff
@@ -40,6 +52,25 @@ pub struct CanisterAExports {
     state: Rc<RefCell<State>>,
 }
 
+impl CanisterBase for CanisterAExports {
+    fn pre_update(&self, _method_name: &str, _method_type: MethodType) {
+        self.update_metrics();
+    }
+}
+
+impl Metrics for CanisterAExports {
```
blutooth, line 61: Let's move this impl into Metrics. Reduce this part to impl Metrics for CanisterAExports {}

ic-canister/tests/canister_b/src/lib.rs
```diff
@@ -89,6 +95,18 @@ impl CanisterB {
             .unwrap();
         true
     }
+
```
blutooth, line 98: Let's move this out of the basic example

blutooth (changes requested):
See comments.

AlessandroRiet (approved):

PR #71, opened 2022-06-27
[CPROD-899] Fix factory upgrade method

Maximkaaa:
This commit fixes two issues:
1. The upgrade method in the factory API used `FactoryCanister::get_canister_bytecode` method. When such method is accessed from the FactoryExport struct (the one that is created just for trait APIs), which doesn't and cannot provide an implementation for this method. Instead, we remove this API method from the trait, leaving the actual method logic in the trait. This way, the factory structure implementing the `FactoryCanister` trait can have a one-line API method implementation that just calls the method in the trait providing the wasm code as an argument.
2. The `pre_upgrade` and `post_upgrade` macros reuse the code for other types of API methods, which included requesting the method call arguments from IC. Such a call resulted in a runtime error in a canister when trying to upgrade. We remove the arguments request for `pre_` and `post_upgrade` methods, and add compile-time check that those methods don't expect such arguments.

Due to the discovery of the first issue, it's now clear that a canister trait actually cannot safely have unimplemented methods (because of how the API methods are exported). This means that the export internal structure will always be an empty impl block, so we can move this struct generation into the `generate_exports` macro, which simplifies the usage of canister traits and prevent errors when writing export structs.

### DIFF
--- a/ic-factory/src/api.rs
+++ b/ic-factory/src/api.rs
@@ -4,6 +4,7 @@ use candid::{Nat, Principal};
 use ic_canister::{generate_exports, query, update, AsyncReturn, Canister};
 
 use ic_helpers::management;
+use ic_helpers::management::WasmModule;
 
 use super::{error::FactoryError, FactoryState};
 
@@ -27,8 +28,6 @@ pub trait FactoryCanister: Canister + Sized {
         use ic_storage::IcStorage;
         FactoryState::get()
     }
-
-    fn get_canister_bytecode(&self) -> Option<Vec<u8>>;
 
     /// Returns the checksum of a wasm module in hex representation.
     #[query(trait = true)]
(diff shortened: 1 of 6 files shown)


ic-factory/src/api.rs
```diff
@@ -66,14 +65,15 @@ pub trait FactoryCanister: Canister + Sized {
 
     /// Upgrades canisters controller by the factory and returns a list of outdated canisters
     /// (in case an upgrade error occurs).
-    #[update(trait = true)]
```
awkure, line 69: It's not a method anymore?

AlessandroRiet (approved):

awkure:
Since we're preventing user for writing exporting struct explicitly, we should also consider doing https://infinityswap.atlassian.net/browse/CPROD-935 as `init` methods for trait canisters are little nontrivial

ic-factory/src/api.rs
awkure: Ok, now I understand the first part :)

awkure (approved):

PR #74, opened 2022-06-30
Recall/restore canister API for the FactoryState

rumatoest:


Maximkaaa (commented):
What is the difference between this new method and the `register` method (except the error messages)? Do we really need both of them?

blutooth:
We want the register function to 
(1) check that the wasm is the smae as the stored wasm
(2) check that the factpry is the sole owner of the canister

rumatoest:
> We want the register function to (1) check that the wasm is the smae as the stored wasm (2) check that the factpry is the sole owner of the canister

What if the wasm it not the same as a stored one? For example it is a previous version. Shouldn't we recall previous versions that we actually own?

Maximkaaa:
Hey, @rumatoest . What's up with this stale PR? Are we still working on this?

rumatoest:
@Maximkaaa Ticket related to this PR was deleted. Thus I guess this PR should be closed as well

Maximkaaa:
Ok, let's close it for now. If we need it later, we can always restore the brunch.

PR #75, opened 2022-07-06
[CPROD-932] Add a function to get candid header for a type

Maximkaaa:


ic-helpers/src/candid_header.rs
```diff
@@ -0,0 +1,65 @@
...
+//! state.
+
+use candid::ser::TypeSerialize;
+use candid::{CandidType, Deserialize};
+use ic_storage::stable::Versioned;
+
+/// Magic prefix used to signify candid encoded binary.
+pub const MAGIC: &[u8] = b"DIDL";
+
+/// Candid header of a versioned state struct.
```
hagsteel, line 11: What is this struct used for?
How is it used?
I think it might be worth documenting these things

hagsteel: This is super nice! Good work

hagsteel (approved):

PR #76, opened 2022-07-14
Fixing buld issue related to candid and clippy checks

rumatoest:
Now it is faling and has to be fixed somehow


hagsteel (approved):

awkure (commented):
What black magic is this?
image

awkure (approved):

PR #77, opened 2022-07-14
[CPROD-964] Refactor the canister factories and add upgrade state check

Maximkaaa:
* Adds update_lock on the state to prevent state modification while
  upgrade or create processes are on the way
* Adds the state check on the canisters before their code is upgraded.
  This state check verifies that the state is not modified or if it is,
  that it uses a new version as defined by the `Versioned` trait.
* Adds the static typed check on factory operation permissions.

### DIFF
--- a/ic-factory/src/api.rs
+++ b/ic-factory/src/api.rs
@@ -1,38 +1,25 @@
+use super::{error::FactoryError, FactoryState};
+use candid::{CandidType, Nat, Principal};
+use ic_canister::{
+    generate_exports, query, update, virtual_canister_call, AsyncReturn, Canister, PreUpdate,
+};
+use ic_cdk::export::candid::utils::ArgumentEncoder;
+use ic_helpers::candid_header::{validate_header, CandidHeader, TypeCheckResult};
+use ic_helpers::management;
+use ic_storage::stable::Versioned;
+use ic_storage::IcStorage;
+use std::collections::HashMap;
 use std::{cell::RefCell, rc::Rc};
 
-use candid::{Nat, Principal};
-use ic_canister::{generate_exports, query, update, AsyncReturn, Canister, PreUpdate};
-
-use ic_helpers::management;
-use ic_helpers::management::WasmModule;
-
-use super::{error::FactoryError, FactoryState};
-
-/// API methods that are added:
-/// * get_checksum
-/// * get_cycles
-/// * top_up
-/// * upgrade
-/// * version
-/// * length
-/// * get_all
-/// * get_icp_fee
-/// * set_icp_fee
-/// * get_icp_to
-/// * set_icp_to
-/// * get_controller
-/// * set_controller
-/// * refund_icp
 pub trait FactoryCanister: Canister + Sized + PreUpdate {
     fn factory_state(&self) -> Rc<RefCell<FactoryState>> {
-        use ic_storage::IcStorage;
         FactoryState::get()
     }
 
     /// Returns the checksum of a wasm module in hex representation.
     #[query(trait = true)]
-    fn get_checksum<'a>(&'a self) -> String {
-        self.factory_state().borrow().factory.checksum.to_string()
+    fn get_checksum(&self) -> Result<String, FactoryError> {
+        Ok(hex::encode(&self.factory_state().borrow().module()?.hash()))
     }
 
     /// Returns the cycles balances.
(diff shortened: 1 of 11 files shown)


Maximkaaa:
You can see how this PR is applied in the factories in this PR: https://github.com/infinity-swap/IS20/pull/99

awkure (approved):

ic-factory/src/api.rs
```diff
@@ -216,6 +328,32 @@ pub trait FactoryCanister: Canister + Sized + PreUpdate {
...
+        Box::pin(async move {
+            let caller = ic_canister::ic_kit::ic::caller();
+            let state_lock = self.factory_state().borrow_mut().lock()?;
+            let actor = self
+                .factory_state()
+                .borrow_mut()
+                .authorize_owner_int(caller)?
+                .drop_canister(canister_id, &state_lock);
+
+            actor.await?;
```
hagsteel, line 342: Just await the future, I don't see the point in the actor here.
If there is one please add a comment as to why.

ic-factory/src/core.rs
```diff
@@ -1,124 +1,37 @@
...
+    wasm_module: Vec<u8>,
+    init_args: T,
+    cycles: u64,
+    controllers: Option<Vec<Principal>>,
+) -> CallResult<Principal> {
+    let settings = CanisterSettings {
+        controllers,
+        compute_allocation: None,
+        memory_allocation: None,
+        freezing_threshold: None,
```
hagsteel, line 17: `CanisterSettings` implements `Default`
```suggestion
        ..Default::default()
```

ic-factory/src/state.rs
```diff
@@ -1,139 +1,455 @@
...
+    ) -> Result<Authorized<'a, Owner>, FactoryError> {
+        if caller == self.configuration.controller {
+            Ok(Authorized::<'a, Owner> {
+                factory: self,
+                phantom: PhantomData::default(),
+            })
+        } else {
+            Err(FactoryError::AccessDenied)
+        }
+    }
```
hagsteel, line 124: Remove the life time specified here, it's not required:

```suggestion
    pub(crate) fn authorize_owner_int(
        &mut self,
        caller: Principal,
    ) -> Result<Authorized<'_, Owner>, FactoryError> {
        if caller == self.configuration.controller {
            Ok(Authorized::<'_, Owner> {
                factory: self,
                phantom: PhantomData::default(),
            })
        } else {
            Err(FactoryError::AccessDenied)
        }
    }
```

ic-factory/src/state.rs
```diff
@@ -1,139 +1,455 @@
...
+
+    /// Checks if the request caller is the factory conctroller (owner).
+    ///
+    /// # Errors
+    ///
+    /// Returns `FactoryError::AccessDenied` if the caller is not the factory controller.
+    pub fn authorize_owner(&mut self) -> Result<Authorized<Owner>, FactoryError> {
+        let caller = ic_canister::ic_kit::ic::caller();
+        self.authorize_owner_int(caller)
```
hagsteel, line 105: I can't see a reason for `authorize_owner_int`, it might as well all be in this function

ic-helpers/src/candid_header.rs
```diff
@@ -45,6 +45,93 @@ pub fn candid_header<T: CandidType + Versioned>() -> CandidHeader {
...
+                    current_version,
+                }
+            } else {
+                TypeCheckResult::Error {
+                    remote_version: remote_header.version,
+                    current_version,
+                    error_message: generate_state_error(remote_header, &historic_header),
+                }
+            }
+        }
```
hagsteel, line 84:
```suggestion
        Some(historic_header) if historic_header == remote_header.header => TypeCheckResult::Ok {
            remote_version: remote_header.version,
            current_version,
        },
        Some(historic_header) => TypeCheckResult::Error {
            remote_version: remote_header.version,
            current_version,
            error_message: generate_state_error(remote_header, &historic_header),
        },
```

ic-helpers/src/candid_header.rs
```diff
@@ -45,6 +45,93 @@ pub fn candid_header<T: CandidType + Versioned>() -> CandidHeader {
...
+}
+
+fn get_type_definition(state_header: &[u8]) -> candid::TypeEnv {
+    use binread::BinRead;
+    use candid::binary_parser::Header;
+    use std::io::Cursor;
+
+    let mut with_magic = vec![];
+    with_magic.append(&mut MAGIC.to_vec());
+    with_magic.append(&mut state_header.to_vec());
```
hagsteel, line 126: Save some heap allocs:
```suggestion
    with_magic.extend(MAGIC);
    with_magic.extend(state_header);
```

ic-helpers/src/candid_header.rs
```diff
@@ -45,6 +45,93 @@ pub fn candid_header<T: CandidType + Versioned>() -> CandidHeader {
...
+
+    let mut reader = Cursor::new(&with_magic);
+    let header = Header::read(&mut reader).unwrap();
+    let (env, _) = header.to_types().unwrap();
```
hagsteel, line 130: no unwraps

hagsteel (changes requested):

ic-factory/src/api.rs
```diff
@@ -63,62 +50,178 @@ pub trait FactoryCanister: Canister + Sized + PreUpdate {
...
+                .hash()
+                .clone();
+            let mut results = HashMap::new();
+            for (canister, _) in state_checks {
+                if state_rc.borrow().canisters()[&canister] == module_hash {
+                    results.insert(canister, UpgradeResult::Noop);
+                    continue;
+                }
+
+                // Set future to local variable to drop the state before the async call
```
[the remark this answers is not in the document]
awkure: This comment does not make sense since it's not dropped before the `.await`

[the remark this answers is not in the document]
hagsteel: That was my thought but I figured maybe I missed something.

ic-factory/src/state.rs
```diff
@@ -1,139 +1,455 @@
...
+        let controller = self.controller();
+
+        consume_provided_cycles_or_icp(caller, ledger, icp_to, icp_fee, controller)
+    }
+}
+
+/// Abstraction to provided compile time checks for factory method access.
+pub struct Authorized<'a, T> {
+    factory: &'a mut FactoryState,
+    phantom: PhantomData<T>,
```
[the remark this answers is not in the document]
awkure: I think the intention was to have a typesafe flag for `Authorized`, e.g. allowing extending it with `User` or similar structs. This will mean that `User` will have the same `factory` field as `Owner`, but a separate API. This'll bothersome since with your proposal `User` and `Owner` will have the similar `factory` fields which should be equal in any case.

ic-factory/src/state.rs
```diff
@@ -102,19 +102,8 @@ impl FactoryState {
...
-    // cannot be called (due to IC limitation). In this case we use this method with the caller,
-    // retrieved before the `await` point. To prevent incorrect usage of this method it is
-    // `pub(crate)` only.
-    pub(crate) fn authorize_owner_int<'a>(
-        &'a mut self,
-        caller: Principal,
-    ) -> Result<Authorized<'a, Owner>, FactoryError> {
         if caller == self.configuration.controller {
-            Ok(Authorized::<'a, Owner> {
+            Ok(Authorized::<Owner> {
```
hagsteel, line 106: `Authorized` has a lifetime so we should always supply one, however since it's elided (because of `&mut self`) we can provide an anonymous lifetime.
This applies to the return type as well
```suggestion
            Ok(Authorized::<'_, Owner> {
```

hagsteel (changes requested):
Good work! Just some minor bits

ic-factory/src/state.rs
awkure: If it follows elision rules, why should we provide an anonymous lifetime then?

ic-factory/src/state.rs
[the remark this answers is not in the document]
hagsteel: Agreed. Then remove it if it's not used anywhere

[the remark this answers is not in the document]
awkure: But maybe it's a good compile time check for restricting calling owner-related functions such as `set_canister_wasm` so that we won't be needed prefixing it with `self.validate_owner()?` or something like that. I think we should leave it as is.

hagsteel (approved):
Looks good. I'll approve it but could you remove the unused import of PhantomData?

PR #85, opened 2022-07-25
Some fixes regarding canister upgrade PR

awkure:


ic-factory/src/state.rs
```diff
@@ -168,6 +185,8 @@ impl FactoryState {
     ) -> Result<impl Future<Output = CallResult<Principal>>, FactoryError> {
         self.check_lock(lock);
 
+        println!("bef");
```
blutooth, line 188: remove println

ic-factory/src/state.rs
```diff
@@ -480,20 +499,34 @@ async fn consume_provided_icp(
     icp_to: Principal,
     icp_fee: u64,
 ) -> Result<(), FactoryError> {
+    println!("www");
```
blutooth, line 502: rm println

blutooth (approved):

PR #87, opened 2022-07-28
[CPROD-1016] Add cycle auction trait

awkure:
Moving out auction stuff out of is20.


ic-auction/src/state.rs
```diff
@@ -0,0 +1,206 @@
...
+    AuctionInfo, BiddingInfo, Cycles, Timestamp,
+};
+
+// Minimum bidding amount is required, for every update call costs cycles, and we want bidding
+// to add cycles rather then to decrease them. 1M is chosen as one ingress call costs 590K cycles.
+pub const MIN_BIDDING_AMOUNT: Cycles = 1_000_000;
+
+//--------------------------------------------------------------------------------------------------
+// Bidding state
+//--------------------------------------------------------------------------------------------------
```
hagsteel, line 19: Love the comment style, but cap it at around 80 chars
```suggestion
//------------------------------------------------------------------------------
// Bidding state
//------------------------------------------------------------------------------
```

ic-auction/src/state.rs
```diff
@@ -0,0 +1,206 @@
...
+            auction_period: 10u64.pow(9) * 60 * 60 * 24, // 1 day
+            cycles_since_auction: 0,
+            bids: HashMap::new(),
+        }
+    }
+}
+
+//--------------------------------------------------------------------------------------------------
+// Auction state
+//--------------------------------------------------------------------------------------------------
```
hagsteel, line 58:
```suggestion
//------------------------------------------------------------------------------
// Auction state
//------------------------------------------------------------------------------
```

ic-auction/src/state.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    pub fn reset_bidding_state(&mut self) {
+        self.bidding_state = BiddingState {
+            fee_ratio: self.get_fee_ratio(),
+            cycles_since_auction: 0,
+            last_auction: ic::time(),
+            bids: HashMap::new(),
+            auction_period: self.bidding_state.auction_period,
+        };
```
hagsteel, line 110:
```suggestion
        self.bidding_state = BiddingState {
            fee_ratio: self.get_fee_ratio(),
            auction_period: self.bidding_state.auction_period,
            ..Default::default()
        };
```

ic-auction/src/state.rs
```diff
@@ -0,0 +1,206 @@
...
+//     // #[test_case(0, 0, 0.0)]
+//     // #[test_case(0, 1000, 0.0)]
+//     // #[test_case(1000, 0, 1.0)]
+//     // #[test_case(1000, 1000, 1.0)]
+//     // #[test_case(1000, 10_000, 0.5)]
+//     // #[test_case(1000, 1_000_000, 0.125)]
+//     // fn fee_ratio_tests(min_cycles: u64, current_cycles: u64, ratio: f64) {
+//     //     assert_eq!(get_fee_ratio(min_cycles, current_cycles), ratio);
+//     // }
+// }
```
hagsteel, line 206: Either bin or fix

ic-auction/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+pub mod api;
+mod state;
+mod types;
+
+pub mod error;
+
+pub use self::state::*;
+pub use self::types::*;
```
hagsteel, line 8: 1. Don't call the module types, it's entirely nonsense and I don't know why we keep insisting on doing it
2. Don't glob imports, especially since there are only two types in there
3. Don't call module types (this is worth mentioning twice). This is like naming a variable `variable`

ic-auction/src/types.rs
```diff
@@ -0,0 +1,40 @@
+use ic_cdk::export::candid::{CandidType, Deserialize};
+use ic_helpers::tokens::Tokens128;
+
+pub type Cycles = u64;
+pub type TxId = u64;
+pub type Timestamp = u64;
+
+#[derive(CandidType, Debug, Clone, Deserialize, PartialEq)]
+pub struct AuctionInfo {
```
hagsteel, line 9: Do we need this type to be public or can it be `pub(crate)` ?
Same for all the fields

ic-auction/src/types.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub tokens_distributed: Tokens128,
+    pub cycles_collected: Cycles,
+    pub fee_ratio: f64,
+    pub first_transaction_id: TxId,
+    pub last_transaction_id: TxId,
+}
+
+/// Current information about upcoming auction and current cycle bids.
+#[derive(CandidType, Debug, Clone, Deserialize)]
+pub struct BiddingInfo {
```
hagsteel, line 21: Same question as the above struct (does it need to be public)

ic-factory/src/api.rs
```diff
@@ -18,6 +18,8 @@ pub trait FactoryCanister: Canister + Sized + PreUpdate {
     // default definitions of the `FactoryCanister` will use `ic-factory`
     // state, instead of the state of the factory canister, that
     // implements this trait.
+    // TODO: remove default implementation since the user may forget
```
hagsteel, line 21: Is there a reason the default impl wasn't removed?

hagsteel (changes requested):
Good work with documentation!

ic-auction/src/api.rs
```diff
@@ -0,0 +1,139 @@
...
+    fn canister_pre_update(&self, method_name: &str, _method_type: ic_canister::MethodType) {
+        if method_name == "run_auction" {
+            if !self.auction_state().borrow().bidding_state.is_auction_due() {
+                ic_cdk::println!("Too early to begin auction");
+            }
+        } else {
+            if let Err(auction_error) = self.run_auction() {
+                ic_cdk::println!("Auction error: {auction_error:#?}");
+            }
+        }
```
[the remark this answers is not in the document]
awkure: Yeah, it's better than not processing the call at all if something fails.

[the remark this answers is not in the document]
hagsteel: How about returning a `Result` then?

ic-auction/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+pub mod api;
+mod state;
+mod types;
+
+pub mod error;
+
+pub use self::state::*;
```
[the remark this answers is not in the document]
awkure: (every)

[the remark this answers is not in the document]
awkure: But yeah, adding a new type to state and forgetting to update this export I think will be more tedious if we'll inherit this practice everywhere

ic-auction/src/state.rs
awkure: Strange, I thought copilot was smart enough setting it to 80 chars

ic-auction/src/lib.rs
awkure: 1. How else you suggest to name it? This is basically a module for domain specific types 🤷 
2. Well, I'll export the module then

ic-auction/src/types.rs
awkure: Yes, because we overwrite `disburse_rewards` as a part of public API

ic-auction/src/types.rs
awkure: I exported it for the is20 to overwrite the endpoints to follow is20 spec

ic-factory/src/api.rs
awkure: Yeah, may break the downstream and I didn't want for this PR to relate to it

ic-auction/src/state.rs
hagsteel: COPILOT!!!!! D:

ic-auction/src/lib.rs
hagsteel: Isn't every module a module for domain specific types in almost every programming project ever?!?!

Can't they live in the same module as state?

I can't stress how poor it is to name something `types`. If we are making a programming language, sure, but just calling a module "types" tells you nothing about what's in there

ic-auction/src/types.rs
hagsteel: yeah if they are needed outside of the crate then all good

ic-factory/src/api.rs
hagsteel: Is there a tracking issue for this so it's not forgotten?

ic-auction/src/lib.rs
awkure: Take is20 for example
https://github.com/infinity-swap/IS20/blob/main/src/token/api/src/types.rs
Would you really like increasing `state.rs` module by additional 2 hundred lines with structs that are used not only in state? I don't really understand what's the problem with having this module though as I never had this confusion myself (probably because of my haskell background).

ic-factory/src/api.rs
awkure: I thought there was 
https://infinityswap.atlassian.net/browse/CPROD-1056

ic-auction/src/api.rs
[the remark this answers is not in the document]
awkure: We don't want unrelated to auction update functions to fail. This basically emulates the process that happens on the background and, for instance, if the user want's to set up a controller to the canister, we should still allow it even if running auctions on the background fails.

Maximkaaa (approved):

rumatoest (approved):
LGTM but I guess that I do not have enough expertise here.

blutooth (approved):

PR #88, opened 2022-08-03
[CPROD-1056] Remove default implementation of state getters in traits

awkure:
Problem: As there's a default `State::get()` call that is used
in a trait canister there's a problem that it's only crate-local.
Hence, the user has to remember to overwrite it or he'll be forced
to face the UB as if there's no stable state yet it'll return
a default state. We should add a macro that has compile-time checks
for state getters and that helps to add a default implementation
of state getter inside `generate_exports!` struct (we don't really
use the state there, only the interface export).


ic-canister/tests/canister_a/src/lib.rs
```diff
@@ -19,9 +19,8 @@ impl Versioned for StateA {
 }
 
 pub trait CanisterA: Canister {
-    fn state(&self) -> Rc<RefCell<StateA>> {
-        StateA::get()
-    }
+    #[state_getter]
```
awkure, line 22: don't really like this name but couldn't come up with a better one

rumatoest (approved):

Maximkaaa (approved):
Adding another magical macro is not ideal, but I agree that it is still better then having magical state appearing from nowhere. The approach I've implemented was too error-prone, and this one has at least some compile-time protection for people who don't read documentation of the tools they use (e.g. all developers).

PR #90, opened 2022-08-09
Cprod 895 stable vec

hagsteel:
Adds stable log and map.
This does not address the max key / value size issue (we'll deal with that in another PR)


aewc (approved):
For me everything is very good

blutooth (approved):

PR #94, opened 2022-08-23
Cprod 895 stable vec

hagsteel:
Adding the multimap struct


aewc (approved):
great, all looks good to me

Maximkaaa (approved):

PR #95, opened 2022-09-05
Add the necessary deriving instances for CandidHeader

awkure:


Maximkaaa (approved):

ic-helpers/src/candid_header.rs
```diff
@@ -21,7 +21,7 @@ pub const MAGIC: &[u8] = b"DIDL";
 ///
 /// This header can be used to transfer information about the type between canisters or to verify
 /// that the type used by a canister is what the consumr expects.
-#[derive(Debug, CandidType, Deserialize)]
+#[derive(Debug, CandidType, Deserialize, Clone, PartialEq)]
```
hagsteel, line 24: Add `Eq` as well, because all underlying types implement `Eq`

```suggestion
#[derive(Debug, CandidType, Deserialize, Clone, PartialEq, Eq)]
```

hagsteel (changes requested):

hagsteel (approved):

PR #96, opened 2022-09-08
[CPROD-1118] factory auto top up cycles

itsyaasir:


ic-factory/src/state.rs
```diff
@@ -487,45 +495,74 @@ fn consume_message_cycles() -> Result<u64, FactoryError> {
...
+    transfer(
+        ledger,
+        icp_fee + TOTAL_FEE,
+        Some((&PrincipalId(caller)).into()),
+        id,
+    )
+    .await?;
+
+    // Tops up the Factory canister with cycles. We only require amount equal to `MIN_CANISTER_CYCLES * 2` + Necessary FEES and send the rest to `icp_to`.
+    let amount = 36_500_000;
```
blutooth, line 534: This should be 5 trillion cycles, it should also be written as a constant and stored at the top of the file.

ic-factory/src/api.rs
```diff
@@ -26,6 +26,7 @@ pub trait FactoryCanister: Canister + Sized + PreUpdate {
     /// otherwise, cycles balances of `principal` is returned.
     /// If `principal` does not exists, `None` is returned.
     #[update(trait = true)]
+    #[allow(clippy::needless_lifetimes)]
```
hagsteel, line 29: Let's not have this. I think @awkure has fixed this so correct life time syntax can be used

ic-factory/src/state.rs
```diff
@@ -487,45 +494,77 @@ fn consume_message_cycles() -> Result<u64, FactoryError> {
...
         .await
         .map_err(FactoryError::LedgerError)?;
 
-    if balance < icp_fee + DEFAULT_TRANSFER_FEE.get_e8s() {
-        return Err(FactoryError::NotEnoughIcp(
-            balance,
-            icp_fee + DEFAULT_TRANSFER_FEE.get_e8s(),
-        ));
-    }
+    // The TOTAL_FEE is 3 times, because of the different transfers :-
```
hagsteel, line 513:
```suggestion
    // The TOTAL_FEE is multiplied by 3 because of the different transfers:
```

ic-factory/src/top_up.rs
```diff
@@ -0,0 +1,100 @@
...
+    let rate = virtual_canister_call!(
+        principal,
+        "get_icp_xdr_conversion_rate",
+        (),
+        IcpXdrConversionRateCertifiedResponse
+    )
+    .await
+    .map_err(|e| FactoryError::GenericError(e.1))?;
+
+    Ok(rate)
```
hagsteel, line 46:
```suggestion
    virtual_canister_call!(
        principal,
        "get_icp_xdr_conversion_rate",
        (),
        IcpXdrConversionRateCertifiedResponse
    )
    .await
    .map_err(|e| FactoryError::GenericError(e.1))
```

ic-factory/src/top_up.rs
```diff
@@ -0,0 +1,100 @@
...
+        minting_canister,
+        "notify_top_up",
+        (notify_details,),
+        Result<u128, NotifyError>
+    )
+    .await
+    .map_err(|e| FactoryError::GenericError(e.1))?
+    .map_err(|e| FactoryError::GenericError(e.to_string()))?;
+
+    Ok(cycles)
```
hagsteel, line 99: This is not strictly necessary.
There is no harm in naming the variable and returning that, but if you want to make it shorter:
```suggestion
    virtual_canister_call!(
        minting_canister,
        "notify_top_up",
        (notify_details,),
        Result<u128, NotifyError>
    )
    .await
    .map_err(|e| FactoryError::GenericError(e.1))?
    .map_err(|e| FactoryError::GenericError(e.to_string()))
```

hagsteel (changes requested):

Maximkaaa (commented):
This PR is not compiling ATM. Please, re-request the review when it's possible to build it.

ic-factory/src/top_up.rs
```diff
@@ -0,0 +1,148 @@
...
+        .await
+        .map_err(|e| FactoryError::LedgerError(e.1))
+}
+
+pub(crate) async fn mint_cycles_to_factory(
+    block_height: BlockHeight,
+) -> Result<u128, FactoryError> {
+    let minting_canister = Principal::from_text(CYCLE_MINTING_CANISTER).expect("const conversion");
+    let to_canister =
+        CanisterId::new(ic_canister::ic_kit::ic::id().into()).expect("const conversion");
```
Maximkaaa, line 80: This is not a `const conversion`

ic-factory/src/state.rs
```diff
@@ -456,76 +461,82 @@ impl Default for FactoryConfiguration {
...
+        .min(icp_fee);
+
+    if balance < top_up_fee + icp_fee {
+        Err(FactoryError::NotEnoughIcp(balance, top_up_fee + icp_fee))?;
     }
 
-    consume_icp(caller, icp_fee, icp_to, ledger).await?;
+    let block_height =
+        top_up::transfer_icp_to_cmc(top_up_fee, ledger, Subaccount::from(&PrincipalId(caller)))
+            .await?;
```
Maximkaaa, line 510: I have a feeling that some kind of error sneaked through here:
* `icp_fee` is the amount of ICP to be sent to the `icp_to` principal just as a commission for the operation
* `top_up_fee` is the amount of ICP to be converted to cycles for the factory
If we don't want to have any extra commission for creating canisters, `icp_fee` can be set to 0. But in that case `top_up_fee` will also be equal to 0, so no cycles will be added to the factory. This will also mean that if `icp_fee` is set below amount corresponding to `MIN_CANISTER_CYCLES`, then the canister will be created with less than `MIN_CANISTER_CYCLES` which sounds strange. It's even possible for the canister creation to fail if the cycles amount is to little in the end.

Please, check that this logic works as intended.

ic-factory/src/top_up.rs
```diff
@@ -0,0 +1,148 @@
...
+
+use crate::error::FactoryError;
+
+const CYCLE_MINTING_CANISTER: &str = "rkp4c-7iaaa-aaaaa-aaaca-cai";
+
+/// This function calculates the amount required for minting cycles for a canister.
+pub async fn cycles_to_icp(cycles: u64) -> Result<u64, FactoryError> {
+    let rate = get_conversion_rate().await?.data;
+
+    let icp_per_xdr = 10_000.0 / rate.xdr_permyriad_per_icp as f64;
```
Maximkaaa, line 22: We don't have to use f64 here (in this whole function, not only this line). If you first do all multiplications and then all divisions, unsigned integers would work just fine. Also, check for division by 0 is necessary here I believe.

ic-factory/src/top_up.rs
```diff
@@ -0,0 +1,148 @@
...
+    let args = SendArgs {
+        memo: MEMO_TOP_UP_CANISTER,
+        amount: Tokens::from_e8s(amount),
+        fee: DEFAULT_TRANSFER_FEE,
+        from_subaccount: Some(caller_subaccount),
+        to,
+        created_at_time: None,
+    };
+
+    virtual_canister_call!(ledger, "send_dfx", (args,), u64)
```
Maximkaaa, line 70: Do not use `send_dfx` method. Use `transfer` instead. `send_dfx` is for internal use and can be changed at any moment.

ic-factory/src/top_up.rs
```diff
@@ -0,0 +1,148 @@
...
+    TOKEN_SUBDIVIDABLE_BY,
+};
+use num_traits::ToPrimitive;
+
+use crate::error::FactoryError;
+
+const CYCLE_MINTING_CANISTER: &str = "rkp4c-7iaaa-aaaaa-aaaca-cai";
+
+/// This function calculates the amount required for minting cycles for a canister.
+pub async fn cycles_to_icp(cycles: u64) -> Result<u64, FactoryError> {
```
Maximkaaa, line 19: It would be nice to move the math to a separate function from here and add a few more unit tests for this math. I would suggest using icp to cycles conversion function from IC code to write tests like `given N ICP: covert N ICP to cycles with IC code, then convert resulting value backwards with this code, check that the result equals N` for different (also extreme) values of N.

Maximkaaa (changes requested):
Please, check out the comments to the code.

Also, I just want to make sure that what this PR does is really what we intend and there was no miscommunication. Originally there were two ways to create a canister:
1. By providing cycles to the factory directly. All those cycles would be used to create the new canister. In this case the creator of the canister can create it with the needed amount of cycles right away.
2. By providing ICP. The amount of ICP required to create a canister is set by the owner with `icp_fee` setting. All these ICP go to the owner (in order for the owner to later convert them to cycles and top up the factory). In this case the canister is always created with set amount of cycles available (~ 9 * 10^11 cycles considering canister creation fee).

After this change we remove the p.1 entirely, so users can only create canisters with ICP and not with cycles directly. Also, the fee for creation is divided into two parts: some amount is converted to cycles (~ 49 * 10^11 cycles will be available in the canister, the amount is increased by 5 times) which are used for canister creation, and some amount is transferred to the owner (`icp_fee` setting).

Is this correct, @blutooth ?

ic-factory/src/top_up.rs
```diff
@@ -0,0 +1,151 @@
...
+    ledger_canister::{
+        AccountIdentifier, BlockHeight, Subaccount, Tokens, TransferArgs, DEFAULT_TRANSFER_FEE,
+        TOKEN_SUBDIVIDABLE_BY,
+    },
+};
+
+use crate::error::FactoryError;
+
+/// This Principal is a slice equivalent to `rkp4c-7iaaa-aaaaa-aaaca-cai`.
+const CYCLES_MINTING_CANISTER: Principal = Principal::from_slice(&[0, 0, 0, 0, 0, 0, 0, 4, 1, 1]);
```
awkure, line 19: We still use a default one? We need to be able to parametrise this principal as we do with ledger in our repos. Could be a separate PR though

awkure (approved):

ic-factory/src/top_up.rs
itsyaasir: We don't actually need to parametrise this, because we will be using the NNS cmc canister to communicate directly hence no need for it.

PR #98, opened 2022-09-12
[CPROD-1165] do not add methods to IDL if no_api feature is used

itsyaasir:
This ticket gets rid of `mint` and `burn` features when `no_api` feature is used. We can not run the AMM pair with `no_api` feature as this will also get rid of the `mint` and `burn` methods

### DIFF
--- a/ic-canister/ic-canister-macros/src/api.rs
+++ b/ic-canister/ic-canister-macros/src/api.rs
@@ -611,7 +611,7 @@ pub(crate) fn generate_idl() -> TokenStream {
                 let mut rets = Vec::new();
                 #(#rets)*
                 let func = Function { args, rets, modes: #modes };
-                if !cfg!(feature = "no_api") || (#name != "burn" && #name != "mint") {
+                if !cfg!(feature = "no_api") {
                     service.push((#name.to_string(), Type::Func(func)));
                 }
             }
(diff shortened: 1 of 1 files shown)


hagsteel (approved):
I'll approve it but it feels a bit like we are building all these magical rules in here that can easily get missed

ic-canister/ic-canister-macros/src/api.rs
```diff
@@ -612,7 +611,9 @@ pub(crate) fn generate_idl() -> TokenStream {
                 let mut rets = Vec::new();
                 #(#rets)*
                 let func = Function { args, rets, modes: #modes };
-                service.push((#name.to_string(), Type::Func(func)));
+                if !cfg!(feature = "no_api") || (#name != "burn" && #name != "mint") {
```
Maximkaaa, line 614: There should not be check for any specific method names here. This crate is not only used for IS20 tokens. And I don't really get why you want to add these conditions in the first place...

Maximkaaa (changes requested):

ic-canister/ic-canister-macros/src/api.rs
itsyaasir: If the feature `no_api` is set, all the methods are excluded.
I will make a separate PR for IS20 to propagate the `no_api` feature.

itsyaasir: @Maximkaaa  This is the IS20 [PR](https://github.com/infinity-swap/IS20/pull/130

Maximkaaa (approved):

PR #100, opened 2022-09-13
[CPROD-1163] Update IC dependency

awkure:
Currently there is a bug when trying to deploy v1-amm:

```
Error: The Replica returned an error: code 5, message: "Wasm module of canister rrkah-fqaaa-aaaaa-aaaaq-cai is not valid: Wasm module has an invalid import section. Module imports function '__wbindgen_describe' from '__wbindgen_placeholder__' that is not exported by the runtime."
``` 

That is because ic devs have inconsistent versioning in their codebase. We've decided to upgrade the ic dependencies to be able to use their canisters further.

### DIFF
--- a/ic-canister/ic-canister-macros/src/api.rs
+++ b/ic-canister/ic-canister-macros/src/api.rs
@@ -104,7 +104,7 @@ pub(crate) fn api_method(
 
     for arg in args {
         let (arg_type, arg_pat) = match arg {
-            FnArg::Receiver(r) => {
+            FnArg::Receiver(_) => {
                 has_self = true;
                 continue;
             }
--- a/ic-canister/ic-canister-macros/src/api.rs
+++ b/ic-canister/ic-canister-macros/src/api.rs
@@ -292,19 +292,20 @@ pub(crate) fn state_getter(_attr: TokenStream, item: TokenStream) -> TokenStream
 
     let segment = path.path.segments.iter().last();
 
-    if segment.is_none() {
-        return syn::Error::new(
-            input.span(),
-            format!(
-                "unexpected return type for state getter: {:#?}",
-                return_type
-            ),
-        )
-        .to_compile_error()
-        .into();
-    }
-
-    let state_type = segment.expect("already checked").ident.to_string();
+    let state_type = match segment {
+        Some(segment) => segment.ident.to_string(),
+        None => {
+            return syn::Error::new(
+                input.span(),
+                format!(
+                    "unexpected return type for state getter: {:#?}",
+                    return_type
+                ),
+            )
+            .to_compile_error()
+            .into()
+        }
+    };
 
     // Check that the body of the getter is empty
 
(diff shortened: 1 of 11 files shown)


ic-factory/src/api.rs
```diff
@@ -26,6 +26,7 @@ pub trait FactoryCanister: Canister + Sized + PreUpdate {
     /// otherwise, cycles balances of `principal` is returned.
     /// If `principal` does not exists, `None` is returned.
     #[update(trait = true)]
+    #[allow(clippy::needless_lifetimes)]
     fn get_cycles<'a>(&'a self, principal: Option<Principal>) -> AsyncReturn<Option<Nat>> {
```
hagsteel, line 30: I thought this was solved, the issue with the life times.
This should be valid to write as 
```rust
fn get_cycles(&self, principal: Option<Principal>) -> AsyncReturn<Option<Nat>>
```
and if `AsyncReturn` has a life time argument it should be written as `AsyncReturn<'_, Option<Nat>>`

Was this never fixed or do we have a regression?

ic-helpers/src/management/canister.rs
```diff
@@ -443,13 +441,16 @@ pub struct CallSignature {
 pub fn der_encode_pub_key(pk: &Pubkey) -> Vec<u8> {
     let pubkey = PublicKey::parse_slice(pk.as_bytes(), None).expect("not a valid public key");
     let pubkey_bytes_uncompress = pubkey.serialize();
-    let der_encoded_public_key: PublicKeyDocument = SubjectPublicKeyInfo {
-        algorithm: Secp256k1::algorithm_identifier(),
+    let der_encoded_public_key: pkcs8::Document = SubjectPublicKeyInfo {
+        algorithm: AlgorithmIdentifier {
+            oid: ObjectIdentifier::new("1.2.840.10045.2.1")
```
hagsteel, line 446: Don't use `expect` here. If we are hell bent on doing this at least use the supplied function `ObjectIdentifier::new_unwrap`

hagsteel (approved):
Approving but please review the comments (the suggestions is just changing error messages to lower case for consistency)

ic-factory/src/api.rs
awkure: 🤷

hagsteel: Alright, leave it for now, we'll make a separate issue for this

ic-canister/ic-canister-macros/src/api.rs
```diff
@@ -102,19 +102,11 @@ pub(crate) fn api_method(
     let mut args_destr = Punctuated::new();
     let mut has_self = false;
 
-    let mut self_lifetime = quote! {};
-
     for arg in args {
         let (arg_type, arg_pat) = match arg {
             FnArg::Receiver(r) => {
```
itsyaasir, line 107:
```suggestion
            FnArg::Receiver(_) => {
```

itsyaasir (approved):
Everything looks good :)

F3kilo (approved):
We can remove one `expect`, but I think it is not necessary.

PR #101, opened 2022-09-16
Refactor utility canister

awkure:
This PR introduces `ic-exports` (for all of the ic deps) and `canister-sdk` (as a wrapper over all of those crates to simplify dependencies in other projects) crates. Also it refactors it a little and removes deprecated code. I'll update the docs in a separate PR.


hagsteel (approved):
There are some clippy warnings but I think they are fixed in another PR so should be fine

F3kilo (changes requested):
A pair of documentation comments give wrong information.

F3kilo (approved):

PR #106, opened 2022-09-22
Add `get_curr_metrics` to the Metrics trait

Maximkaaa:
Adds a method to get only the current metrics value for all implementers of the Metrics canister trait.

It also adds history length limit for the metrics history (1 year by default).


Cargo.toml
```diff
@@ -3,7 +3,6 @@ members = [
   "ic-factory",
   "ic-auction",
   "ic-helpers",
-  "ic-stable-storage",
```
awkure, line 6: intentional?

awkure (approved):

Cargo.toml
Maximkaaa: Good thing you asked about this. I've completely forgot about this change. But my question is, is this crate supposed to be included? `cargo check` fails with it, because it has wrong `ic` dependency. Did someone forget something when he did something?

hagsteel (changes requested):
Looks fine but fix the merge conflict

ic-metrics/src/map.rs
```diff
@@ -10,14 +10,19 @@ pub struct MetricsStorage {
     pub metrics: MetricsMap<MetricsData>,
 }
 
-#[derive(CandidType, Deserialize, IcStorage, Default, Clone, Debug)]
+#[derive(CandidType, Deserialize, IcStorage, Default, Clone, Debug, PartialEq)]
```
hagsteel, line 13:
```suggestion
#[derive(CandidType, Deserialize, IcStorage, Default, Clone, Debug, PartialEq, Eq)]
```

hagsteel (changes requested):
Some clippy warnings in there

hagsteel (approved):
:1st_place_medal:

PR #108, opened 2022-09-27
Visibility fix in api method macro

awkure:
Something that triggers clippy with unused methods


itsyaasir (approved):

hagsteel:
My bad, we did not need this

PR #120, opened 2022-10-24
update readme and add more info

itsyaasir:


Cargo.toml
```diff
@@ -1,19 +1,19 @@
...
-  "ic-canister/tests/canister-a",
-  "ic-canister/tests/canister-b",
-  "ic-canister/tests/canister-c",
-  "ic-canister/tests/canister-d",
+    "ic-exports",
+    "ic-factory",
+    "ic-auction",
+    "ic-metrics",
+    "ic-helpers",
+    "ic-stable-storage",
```
F3kilo: There is no `ic-stable-storage` package in the workspace now. But the `ic-stable-structures` is present.

F3kilo (changes requested):

Cargo.toml
F3kilo: Already fixed)

F3kilo (approved):

Maximkaaa (approved):

PR #122, opened 2022-11-04
Add `mark_canister` function

Maximkaaa:


itsyaasir (approved):

Maximkaaa:
We figured out a better way of doing this (using `#[no_mangle]`

PR #133, opened 2023-01-05
[CPROD-1364] Correct factory cycles consumption + cmc mock canister

Maximkaaa:
This PR is the first step to allow all the users to create pairs through the factory. There are a few changes here:
* Fix the amount of cycles consumed by a factory when creating a canister. New values are to ensure that after creating a canister the factory has more cycles than it had before, to prevent cycles draining through spamming new canisters.
* Add methods to change CMC canister that the factory uses. These methods are supposed to be used only for testing, since the real CMC canister can only have the hard-coded principal, which is set to the factory by default.
* Add cmc-mock canister crate for testing cycles calculations when factory creates a canister.


itsyaasir (approved):
LGTM

PR #139, opened 2023-02-24
Fix tokens256 from f64 conversion

Maximkaaa:


itsyaasir:
This function `tokens256_from_nat` in the `tokens.rs` has an error, and I have no idea why the CI passed. it's just import error though @Maximkaaa

itsyaasir (approved):

PR #140, opened 2023-03-01
[CPROD-1541] ICRC-1 token payment bridge for canisters

Maximkaaa:
Adds payment terminal (bridge) for safe transactions with ICRC-1 canisters.


ic-payments/src/token_terminal.rs
```diff
@@ -0,0 +1,455 @@
...
+/// sure that the transaction exists.
+pub const UNKNOWN_TX_ID: u128 = u64::MAX as u128;
+
+/// Default number of retries in case of IC error, before a transfer stored into the list for
+/// recovery.
+const N_RETRIES: usize = 3;
+
+/// Default period when deduplication of a transaction is possible. This is set by the token
+/// implementation. 24 hours used here is the most common value, used by ICP and SNS-1 ledgers.
+const DEFAULT_DEDUP_PERIOD: u64 = 10u64.pow(9) * 60 * 60 * 24;
```
itsyaasir, line 27: Don't let @hagsteel see this. Since it's time related value, I recommend we use `Duration`, same with `TX_WINDOW`, more clear.

I think we can call `..as_nanos()` to be used in the `deduplication_period`
```suggestion
const DEFAULT_DEDUP_PERIOD: Duration = Duration::from_secs(60 * 60 * 24);
```

ic-payments/src/balances.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    /// Decrease the `account_owners`'s balance by the given `amount`.
+    fn debit(
+        &mut self,
+        account_owner: Principal,
+        amount: Tokens128,
+    ) -> Result<Tokens128, BalanceError>;
+}
+
+impl<T: Balances> Balances for Rc<RefCell<T>> {
```
blutooth, line 35: should this be a phantom type?

ic-payments/src/error.rs
```diff
@@ -0,0 +1,178 @@
...
+
+use crate::BalanceError;
+
+pub type Result<T> = std::result::Result<T, InternalPaymentError>;
+
+/// Reason for a transfer to fail
+#[derive(Debug, PartialEq, CandidType, Deserialize, Error)]
+pub enum TransferFailReason {
+    #[error("token canister does not exist or doesn't follow the ICRC-1 standard")]
+    NotFound,
```
blutooth, line 15: MethodNotFound ?

ic-payments/src/balances.rs
Maximkaaa: Yes, this provides implementation for a rc of any `Balances` type to simplify usage for in-memory states.

ic-payments/src/error.rs
Maximkaaa: This can by either method or the canister itself. And we cannot distinguish these cases except if we try to parse string error message.

ic-payments/src/token_terminal.rs
Maximkaaa: Actually, I disagree with this position. In IC all time-related staff is u64, nowhere in there libs or internals they use Duration. So I'd rather stick to what the framework is using. @hagsteel

ic-payments/src/error.rs
```diff
@@ -0,0 +1,178 @@
...
+            | TransferError::TemporarilyUnavailable
+            | TransferError::Duplicate { .. }
+            | TransferError::GenericError { .. } => {
+                Self::TransferFailed(TransferFailReason::Rejected(err))
+            }
+            TransferError::BadFee { expected_fee } => {
+                Self::WrongFee(Tokens128::from_nat(&expected_fee).unwrap_or(Tokens128::MAX))
+            }
+        }
+    }
```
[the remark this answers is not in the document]
Maximkaaa: I wrote it explicitly on purpose. It is critical that we handle every error case in a correct way, so listing all possible values and show exactly how which value is handled is better IMO.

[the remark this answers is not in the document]
itsyaasir: Ok. well my thinking was since you are catching all err variants and handling with the same error, but yeah we can also do this.

ic-payments/src/token_terminal.rs
Maximkaaa: Because API returns u64 and takes u64. Everyone who worked with IC would expect u64. Candid doesn't support Duration. Using Duration in this case just adds unnecessary complexity without solid benefit.

itsyaasir (approved):

F3kilo (approved):
Just one suggestion about stable storage.

ic-payments/src/recovery_list.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    fn from_bytes(bytes: Cow<'_, [u8]>) -> Self {
+        Self(candid::decode_one(&bytes).expect("deserialization of transfer failed"))
+    }
+}
+
+/// The only variable size part of the transfer is memo, which is usually 32 bytes. We use 60 bytes
+/// value to account for candid header.
+const VALUE_SIZE_OFFSET: usize = 60;
```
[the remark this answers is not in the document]
Maximkaaa: Thank for suggestion. Applied it.

PR #143, opened 2023-04-07
Fix `Tokens128` conversion

itsyaasir:


ic-helpers/src/tokens.rs
```diff
@@ -238,14 +241,19 @@ impl Tokens256 {
     /// Number of bytes needed to represent the value.
     pub const BYTE_LENGTH: usize = 256 / 8;
 
-    /// Converts the value to Tokens128. Returns `None` if the value is greater than `Tokens128::MAX`.
+    /// Converts the value to Tokens128. Returns `None` if the value is greater
+    /// than `Tokens128::MAX`.
     pub fn to_tokens128(&self) -> Option<Tokens128> {
```
ufoscout, line 246: What issue does this fix? Could you please a test that reproduces the existing bug and shows this fixes it?

ufoscout (changes requested):

ic-helpers/src/tokens.rs
```diff
@@ -238,14 +241,19 @@ impl Tokens256 {
...
-        let limbs = self.0.limbs();
-        if limbs[2].0 != 0 || limbs[3].0 != 0 {
+        let bytes = self.0.to_le_bytes();
+        let mut num = 0;
+        for (i, &byte) in bytes.iter().enumerate().take(Tokens128::BYTE_LENGTH) {
+            num += byte as u128 * 256u128.pow(i as u32);
+        }
+
+        if bytes[16..].iter().any(|&byte| byte != 0) {
```
Maximkaaa, line 253: Replace `16` with `Tokens128::BYTE_LENGTH`, and move the check before the calculation of `num` (we don't need to do all the calculations in case this check fails.

Maximkaaa (changes requested):
Hey, Yasir. Nice fix. One little thing to improve.

ic-helpers/src/tokens.rs
Maximkaaa: Hey, Francesco. There's already test that fails before this fix, but only in x32 environment. Unfortunately, it's not that trivial to run the tests with `wasm32` target in our CI, so I suggest to deal with that issue in a separate PR.

Maximkaaa (approved):

ic-helpers/src/tokens.rs
ufoscout: @Maximkaaa we don't need to run the tests in wasm32, to reproduce the bug it is enough to execute them with  `--target=i686-unknown-linux-gnu`

ufoscout (approved):

PR #164, opened 2023-07-18
Add ic-client

ufoscout:
Add generic Ic client. Code moved from bitfinity-evm-sdk


itsyaasir (approved):

veeso (changes requested):
can we fix just this lint here ic-canister/ic-canister-macros/src/api.rs line 57 this `.filter_map` can be written more simply using `.map`.

Why isn't the CI returning errors for lint on PRs?

veeso (approved):

PR #178, opened 2023-09-27
[EPROD-539] State machine client

F3kilo:
Issue link:
https://infinityswap.atlassian.net/browse/EPROD-539

Implementation of `CanisterClient` trait for state machine framework.

### DIFF
--- a/ic-canister-client/Cargo.toml
+++ b/ic-canister-client/Cargo.toml
@@ -9,6 +9,7 @@ description = "Client for interacting with an IC Canister"
 [features]
 default = []
 ic-agent-client = ["dep:ic-agent"]
+state-machine-tests-client = ["dep:tokio"]
 
 [dependencies]
 candid = { workspace = true }
@@ -19,3 +20,4 @@ serde_json = { workspace = true }
 serde = { workspace = true }
 thiserror = { workspace = true }
 async-trait = { workspace = true }
+tokio = { workspace = true, optional = true, features = ["sync"] }
--- a/ic-canister-client/src/error.rs
+++ b/ic-canister-client/src/error.rs
@@ -1,4 +1,5 @@
 use ic_exports::ic_cdk::api::call::RejectionCode;
+use ic_exports::ic_test_state_machine::UserError;
 use thiserror::Error;
 
 #[derive(Debug, Error)]
@@ -12,6 +13,17 @@ pub enum CanisterClientError {
     #[cfg(feature = "ic-agent-client")]
     #[error("ic agent error: {0}")]
     IcAgentError(#[from] ic_agent::agent::AgentError),
+
+    #[cfg(feature = "state-machine-tests-client")]
+    #[error("state machine test error: {0}")]
+    StateMachineTestError(UserError),
+}
+
+#[cfg(feature = "state-machine-tests-client")]
+impl From<UserError> for CanisterClientError {
+    fn from(error: UserError) -> Self {
+        CanisterClientError::StateMachineTestError(error)
+    }
 }
 
 pub type CanisterClientResult<T> = Result<T, CanisterClientError>;
--- a/ic-canister-client/src/state_machine_tests.rs
+++ b/ic-canister-client/src/state_machine_tests.rs
@@ -1,0 +1,121 @@
+use std::sync::Arc;
+
+use candid::utils::ArgumentEncoder;
+use candid::{CandidType, Decode, Principal};
+use ic_exports::ic_kit::RejectionCode;
+use ic_exports::ic_test_state_machine::{StateMachine, WasmResult};
+use serde::Deserialize;
+use tokio::sync::Mutex;
+
+use crate::{CanisterClient, CanisterClientError, CanisterClientResult};
+
+/// A client for interacting with a canister inside dfinity's
+/// state machine tests framework.
+#[derive(Clone)]
+pub struct StateMachineCanisterClient {
+    state_machine: Arc<Mutex<StateMachine>>,
+    canister: Principal,
+    caller: Principal,
+}
+
+impl StateMachineCanisterClient {
+    /// Creates a new instance of a StateMachineCanisterClient.
+    pub fn new(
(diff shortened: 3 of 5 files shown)


GraDKh (approved):

ic-canister-client/Cargo.toml
```diff
@@ -19,3 +20,4 @@ serde_json = { workspace = true }
 serde = { workspace = true }
 thiserror = { workspace = true }
 async-trait = { workspace = true }
+tokio = { workspace = true, optional = true, features = ["sync"] }
```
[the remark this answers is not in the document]
F3kilo: Done

ic-canister-client/src/error.rs
```diff
@@ -12,6 +13,17 @@ pub enum CanisterClientError {
     #[cfg(feature = "ic-agent-client")]
     #[error("ic agent error: {0}")]
     IcAgentError(#[from] ic_agent::agent::AgentError),
+
+    #[cfg(feature = "state-machine-tests-client")]
+    #[error("state machine test error: {0}")]
+    StateMachineTestError(UserError),
+}
+
+#[cfg(feature = "state-machine-tests-client")]
```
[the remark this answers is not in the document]
F3kilo: The bound `UserError: Error` is not satisfied.

ic-canister-client/src/error.rs
```diff
@@ -1,4 +1,5 @@
 use ic_exports::ic_cdk::api::call::RejectionCode;
+use ic_exports::ic_test_state_machine::UserError;
```
itsyaasir, line 2: This brings clippy warning since the import is used under a feature, We can just use full path in the `StatemachineTestError` variant

ic-canister-client/src/state_machine_tests.rs
itsyaasir, line 1: The name of the module is rather confusing, from the name I will assume it's test module, maybe we can just call it `state_machine.rs`

itsyaasir (approved):
LGTM 👍🏾

ic-canister-client/src/state_machine_tests.rs
F3kilo: The `state_machene` name looks too generic to me. **State machine** term is more known as a behavioural pattern than IC testing framework. So, I'll keep the current name as is.

ic-canister-client/src/error.rs
F3kilo: It used in three places, so I added feature condition to the `use ...` itself.

ic-canister-client/src/error.rs
```diff
@@ -1,4 +1,6 @@
 use ic_exports::ic_cdk::api::call::RejectionCode;
+#[cfg(feature = "state-machine-tests-client")]
+use ic_exports::ic_test_state_machine::UserError;
```
ufoscout, line 3: please do not use conditional `use`, instead, use the full qualified import where UserError is used

ic-helpers/src/tokens.rs
```diff
@@ -264,7 +264,7 @@ impl Tokens256 {
 
     /// Rounded square root of the value.
     pub fn sqrt(&self) -> Self {
-        Self(self.0.sqrt_vartime())
+        Self(self.0.sqrt())
```
ufoscout, line 267: This should be reverted. If you have a compilation error remove the Cargo.lock

ic-canister-client/src/state_machine_tests.rs
```diff
@@ -0,0 +1,121 @@
...
+pub struct StateMachineCanisterClient {
+    state_machine: Arc<Mutex<StateMachine>>,
+    canister: Principal,
+    caller: Principal,
+}
+
+impl StateMachineCanisterClient {
+    /// Creates a new instance of a StateMachineCanisterClient.
+    pub fn new(
+        state_machine: Arc<Mutex<StateMachine>>,
```
ufoscout, line 24: we do you need an async mutex?

ufoscout (changes requested):

ic-canister-client/src/state_machine_tests.rs
F3kilo: Async mutex needed because the client provides async API.

ic-helpers/src/tokens.rs
F3kilo: Done

ic-canister-client/src/error.rs
F3kilo: Done.
Is it part of our code style rules?

ufoscout: no, it's only an attempt to reduce the number of `#[cfg(feature = "???")]` that sprout like mushrooms on a rainy day ;)

ic-canister-client/src/state_machine_tests.rs
ufoscout: maybe I am wrong but I remember that StateMachine client methods are not mutable, so `Arc<StateMachine>` should be enough

F3kilo: StateMachine is !Sync.

ufoscout: I don't get it. Could you please add a simple integration tests that uses the state machine client?

ic-canister-client/src/state_machine_tests.rs
```diff
@@ -0,0 +1,121 @@
...
+        self.state_machine.as_ref()
+    }
+
+    /// Calls a method on the canister.
+    async fn with_state_machine<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce(&StateMachine) -> R,
+    {
+        let state_machine = self.state_machine.lock().await;
+        f(&*state_machine)
```
ufoscout, line 66: this is a blocking call. execute with `tokio::task::spawn_blocking`

F3kilo: Done

ufoscout (approved):

PR #185, opened 2023-10-23
Refactor memory manager

ufoscout:


veeso (approved):

ic-stable-structures/src/structure/common/ring_buffer.rs
```diff
@@ -112,19 +111,35 @@ impl Storable for StableRingBufferIndices {
 }
 
 /// Stable ring buffer implementation
-#[derive(Debug)]
-pub struct StableRingBuffer<T: Storable + Clone + 'static, M: Memory + 'static> {
+pub struct StableRingBuffer<T: Storable + Clone, DataMemory: Memory, IndicesMemory: Memory> {
```
GraDKh, line 114: Do we have a case when we need different memory type for data and indices?
Maybe we can at least specify a default value (`IndicesMemory = DataMemory`)

GraDKh (approved):

ic-stable-structures/src/structure/common/ring_buffer.rs
ufoscout: good point. Fixed

PR #187, opened 2023-10-27
IC stable structures can be Send

ufoscout:


ic-stable-structures/src/structure/cache/btreemap.rs
```diff
@@ -1,26 +1,25 @@
...
+use mini_moka::sync::{Cache, CacheBuilder};
 
 use crate::structure::*;
 
 /// A LRU Cache for StableBTreeMap
 pub struct CachedStableBTreeMap<K, V, M>
 where
-    K: Storable + Clone + Hash + Eq + PartialEq + Ord,
-    V: Storable + Clone,
+    K: Storable + Clone + Send + Sync + 'static + Hash + Eq + PartialEq + Ord,
```
GraDKh, line 11: Do we need to add `Send` and `Sync` constraints here and in the other places?
Since this is a library user may not need the collection to be `Send`.

ic-stable-structures/Cargo.toml
```diff
@@ -7,13 +7,15 @@ edition.workspace = true
...
 mini-moka = { workspace = true }
+parking_lot = { workspace = true }
 thiserror = { workspace = true }
 
 [dev-dependencies]
 anyhow = "1"
 candid = { workspace = true }
 criterion = { workspace = true }
 did = { path = "./tests/did" }
+futures = { workspace = true }
```
GraDKh, line 18: Is it used somewhere?

GraDKh (approved):

ic-stable-structures/Cargo.toml
ufoscout: removed

ic-stable-structures/src/structure/cache/btreemap.rs
ufoscout: They are needed by the internal cache. Anyway, ic_stable_structure entries must be deserialized and this forecs the objects to be send/sync/static even without explicitly requiring it. So, from a user perspective, this changes nothing.
BTW, we need to remove moka, so this could change.

veeso (approved):

PR #188, opened 2023-10-30
Add save copy functionality to memory manager

GraDKh:


ic-stable-structures/src/memory_mapped_files/memory.rs
```diff
@@ -12,56 +15,87 @@ const WASM_PAGE_SIZE_IN_BYTES: u64 = 65536;
...
             base_path,
             is_persistent,
+            created_memory_resources: Default::default(),
+        }
+    }
+
+    pub fn flush_and_save_copies_to(&self, path: impl AsRef<Path>) -> Result<(), MemMapError> {
+        let created_memory_resources = self.created_memory_resources.read();
+        for (file_path, memory) in created_memory_resources.iter() {
+            let file_name = file_path
```
veeso, line 33: maybe here you can direectly convert file_name to str

GraDKh: I need to extract the name of the file first, so it's more convenient to do it first and then convert it to string

ic-stable-structures/src/memory_mapped_files/memory.rs
```diff
@@ -12,56 +15,87 @@ const WASM_PAGE_SIZE_IN_BYTES: u64 = 65536;
...
 impl MemoryMappedFileMemoryManager {
     pub fn new(base_path: PathBuf, is_persistent: bool) -> Self {
         Self {
             base_path,
             is_persistent,
+            created_memory_resources: Default::default(),
+        }
+    }
+
+    pub fn flush_and_save_copies_to(&self, path: impl AsRef<Path>) -> Result<(), MemMapError> {
```
ufoscout, line 30: A description

ic-stable-structures/src/memory_mapped_files/memory.rs
```diff
@@ -12,56 +15,87 @@ const WASM_PAGE_SIZE_IN_BYTES: u64 = 65536;
...
+                file_name
+                    .to_str()
+                    .ok_or(MemMapError::InvalidSourceFileName)?,
+            );
+
+            memory.save_copy(new_path)?;
+        }
+
+        Ok(())
+    }
```
ufoscout, line 46: where is the `flush` part performed?

ic-stable-structures/src/memory_mapped_files/memory.rs
```diff
@@ -12,56 +15,87 @@ const WASM_PAGE_SIZE_IN_BYTES: u64 = 65536;
...
     pub fn new(base_path: PathBuf, is_persistent: bool) -> Self {
         Self {
             base_path,
             is_persistent,
+            created_memory_resources: Default::default(),
+        }
+    }
+
+    pub fn flush_and_save_copies_to(&self, path: impl AsRef<Path>) -> Result<(), MemMapError> {
+        let created_memory_resources = self.created_memory_resources.read();
```
ufoscout, line 31: This implementation allows changing other memory files while the copy one memory is happening. Is this intentional? How do we guarantee consistency while copying?

ufoscout (changes requested):

ic-stable-structures/src/memory_mapped_files/memory.rs
GraDKh: It happens inside `MemoryMappedFile::save_copy`

ic-stable-structures/src/memory_mapped_files/memory.rs
GraDKh: Applied

ic-stable-structures/src/memory_mapped_files/memory.rs
ufoscout: So, two things:
1. Why do you have to explicitly state it in the name of the function?
2. I guess we need a dedicated "flush_all" function

ic-stable-structures/src/memory_mapped_files/memory.rs
ufoscout: That's not true, there are many patterns to lock all the internal memories or to reject write operations while copying

GraDKh: I don't think we should lock on the level of the memory. Because without lock on the state level this lock would be useless - we may start some state operation, do `write_1`, release lock and start creating backup when having some intermediate state.

ufoscout (approved):

F3kilo (approved):

PR #189, opened 2023-11-01
Eprod 562 add endpoints to modify state

GraDKh:


veeso (approved):

ic-stable-structures/src/structure/cache/unbounded.rs
```diff
@@ -33,6 +35,17 @@ where
             cache: SyncLruCache::new(max_cache_items),
         }
     }
+
+    /// Iterator for all stored key-value pairs.
+    pub fn iter(&self) -> StableUnboundedIter<'_, K, V, M> {
```
ufoscout, line 40: The iterator was omitted from the cache because it bypasses the cache completely

ufoscout (changes requested):
check for an alternative to the cache iterator

ic-stable-structures/src/structure/cache/unbounded.rs
GraDKh: Yes it does. But:
1. I don't think there is a way to implement iterator that uses cached values to speedup the iteration.
2. LRU cache won't speedup the iteration over the whole collection by it's nature. Moreover in case when we put the values over which we are iterating to the cache we most probably would just fill it with irrelevant values.

I propose either to leave the implementation as is or just expose `inner` so that it can be iterated explicitly.

GraDKh: Replaced with `inner()` method

ic-stable-structures/src/structure/cache/btreemap.rs
```diff
@@ -33,6 +33,11 @@ where
             cache: SyncLruCache::new(max_cache_items),
         }
     }
+
+    /// Returns the inner collection
```
ufoscout, line 37: even if obvious, could you please add a comment to state that accessing the inner collection will bypass the cache?

ufoscout (approved):

ic-stable-structures/src/structure/cache/btreemap.rs
GraDKh: Added

itsyaasir (approved):

F3kilo (approved):

PR #192, opened 2023-11-09
Client should be clone

ufoscout:


ic-canister-client/src/state_machine_tests.rs
```diff
@@ -139,3 +136,22 @@ impl CanisterClient for StateMachineCanisterClient {
...
 }
+
+#[async_trait::async_trait]
+impl CanisterClient for StateMachineCanisterClient {
+    async fn update<T, R>(&self, method: &str, args: T) -> CanisterClientResult<R>
+    where
+        T: ArgumentEncoder + Send + Sync,
+        R: for<'de> Deserialize<'de> + CandidType,
+    {
+        StateMachineCanisterClient::update(self, method, args).await
```
ufoscout, line 147: The `update` and `query` methods on the `StateMachineCanisterClient` struct are Sync and Send, on the contrary, the ones on the `async_trait` Trait are only Send. 
I cannot make Sync the ones in the Trait because the IcAgent implementation is not Sync

GraDKh (approved):

ic-canister-client/src/state_machine_tests.rs
F3kilo: Usually, a combination of `T: Send + Clone` is enough (like in `std::mpsc::Sender`). Can we require this from the `CanisterClient` trait? This will allow us to make `EvmClient<impl CanisterClient>` clonable too. So, we will be able to clone it in generic code.

F3kilo (approved):

ic-canister-client/src/state_machine_tests.rs
F3kilo: Or, to keep support of non-clonable client impls, we can add `ClonableCanisterClient: CanisterClient + Clone` and implement it on all client impls we currently have.
I think it will be useful, because in all places, where I used the generic client, I thought about changing it to `CanisterClient + Clone`.

itsyaasir (approved):

PR #198, opened 2023-12-20
[Eprod-659] Change logs take logic

ufoscout:


ic-log/src/writer.rs
```diff
@@ -46,43 +48,85 @@ const INIT_LOG_CAPACITY: usize = 128;
...
         LOG_RECORDS.with(|records| {
-            let mut records = records.borrow_mut();
-            let mut result = Vec::with_capacity(max_count);
-            for _ in 0..max_count {
-                if let Some(s) = records.dequeue() {
-                    result.push(s);
+            let records = records.borrow_mut();
+            let all_logs_count = records.0;
+
+            if (from_offset >= all_logs_count) || all_logs_count == 0 {
```
veeso, line 87: unnecessary parenthesis

ic-log/src/writer.rs
```diff
@@ -46,43 +48,85 @@ const INIT_LOG_CAPACITY: usize = 128;
...
+                Logs {
+                    all_logs_count,
+                    logs: vec![],
+                }
+            } else {
+                let mut result = Vec::with_capacity(max_count);
+
+                let from_offset = if (from_offset + records.1.len()) < all_logs_count {
```
veeso, line 95: maybe I would move this block to a separate function and test it standalone, because it's a little hard to read

PR #201, opened 2024-01-11
Add reserved length setting

GraDKh:
Before we were reserving 1 TB of virtual memory space for each memory-mapped file. Which is OK for a single x64 application which has 2^48/2 = 128 TB user address space. But if we are running EVMC tests concurrently where several dozens of files are created for each state we may end up getting memory allocation error (which is actually address space allocation error).
That's why I've added a possibility to change the default reserved length, now it is used in tests only.

### DIFF
--- a/ic-stable-structures/src/memory_mapped_files/memory.rs
+++ b/ic-stable-structures/src/memory_mapped_files/memory.rs
@@ -12,11 +12,17 @@ use crate::memory::MemoryManager;
 
 const WASM_PAGE_SIZE_IN_BYTES: u64 = 65536;
 
+/// When creating mapping we reserve at once 1 TB of address space.
+/// This doesn't allocate any resources (except of address space which is not a problem for x64)
+/// but allows skip remapping/flushing when the file size grows.
+const DEFAULT_MEM_MAP_RESERVED_LENGTH: u64 = 1 << 40;
+
 /// Memory manager that uses one memory mapped filer per one memory id.
 pub struct MemoryMappedFileMemoryManager {
     base_path: PathBuf,
     is_persistent: bool,
     created_memory_resources: RwLock<BTreeMap<PathBuf, MemoryMappedFileMemory>>,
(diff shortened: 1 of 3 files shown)


ic-stable-structures/src/memory_mapped_files/memory_mapped_file.rs
```diff
@@ -170,7 +167,7 @@ mod tests {
     #[test]
     fn should_create_flush_memory_file() {
         with_temp_file(|path| {
-            let mut file_memory = MemoryMappedFile::new(path, true).unwrap();
+            let mut file_memory = MemoryMappedFile::new(path, PAGE_SIZE, true).unwrap();
```
ufoscout, line 170: here, you are using the PAGE_SIZE as file size, I know that this is only a test, but this is confusing. Often people look at the tests to determine how to use something

ufoscout (changes requested):

ic-stable-structures/src/memory_mapped_files/memory.rs
```diff
@@ -27,9 +33,17 @@ impl MemoryMappedFileMemoryManager {
             base_path,
             is_persistent,
             created_memory_resources: Default::default(),
+            file_reserved_length: DEFAULT_MEM_MAP_RESERVED_LENGTH,
         }
     }
 
+    /// Set reserved length for each memory-mapped file.
```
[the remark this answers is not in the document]
GraDKh: I've changed the method name and the description

ic-stable-structures/src/memory_mapped_files/memory_mapped_file.rs
GraDKh: I've extracted DEFAULT_MAX_SIZE constant and added description why PAGE_SIZE is the minimal acceptable vallue

ufoscout (approved):

veeso (approved):

PR #203, opened 2024-01-26
[EPROD-668] Get chunks number from UnboundedMap

F3kilo:
https://infinityswap.atlassian.net/browse/EPROD-668

### DIFF
--- a/ic-stable-structures/src/structure/stable_storage/unbounded.rs
+++ b/ic-stable-structures/src/structure/stable_storage/unbounded.rs
@@ -62,6 +62,20 @@ where
         StableUnboundedIter(self.inner.iter().peekable())
     }
 
+    /// Returns number of chunks, used to store the entity.
+    pub fn chunks_number_of(&self, key: &K) -> Option<ChunkIndex> {
+        let first_chunk_key = Key::new(key);
+        let max_chunk_key = first_chunk_key.clone().with_max_chunk_index();
+
+        // first item in this iter will be the last chunk of the key.
+        self.inner
+            .iter_upper_bound(&max_chunk_key)
+            .filter_map(|(k, _)| {
+                (k.prefix() == max_chunk_key.prefix()).then_some(k.chunk_index() + 1)
+            })
+            .next()
+    }
+
     /// Returns an iterator pointing to the first element below the given bound.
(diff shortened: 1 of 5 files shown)


ic-stable-structures/src/structure/stable_storage/unbounded.rs
```diff
@@ -62,6 +62,20 @@ where
         StableUnboundedIter(self.inner.iter().peekable())
     }
 
+    /// Returns number of chunks, used to store the entity.
+    pub fn chunks_number_of(&self, key: &K) -> Option<ChunkIndex> {
+        let first_chunk_key = Key::new(key);
+        let max_chunk_key = first_chunk_key.clone().with_max_chunk_index();
+
+        // first item in this iter will be the last chunk of the key.
+        self.inner
```
ufoscout, line 71: How much expensive is this? It seems to me that this requires to decode the entire data

ufoscout (changes requested):

ic-stable-structures/src/structure/stable_storage/unbounded.rs
F3kilo: It encodes the `key`, and performs one binary search in the map. The search requires reading (without decoding)  log_2(N) keys.

veeso (approved):

ufoscout (commented):
Waiting for the changes we discussed offline to be implemented

ufoscout (changes requested):
Waiting for the changes we discussed offline to be implemented

GraDKh (approved):

F3kilo:
> Waiting for the changes we discussed offline to be implemented

Already done. There is the `total_chunks_number()` method. 
I'll keep the `chunks_number_of()` method too. It could be useful in another task.

ufoscout:
> I'll keep the `chunks_number_of()` method too. It could be useful in another task.

@F3kilo I would not keep it instead, it has hidden performance implications which are the reason why we implemented the alternative solution.

ufoscout (changes requested):
See my comments

F3kilo:
> @F3kilo I would not keep it instead, it has hidden performance implications which are the reason why we implemented the alternative solution.

Done

PR #207, opened 2024-02-14
use pocket_ic from crates.io

ufoscout:


Cargo.toml
```diff
@@ -75,15 +75,11 @@ tokio-test = "0.4"
...
 icrc-ledger-types = "0.1.0"
 ic-ledger-types = "0.9"
-dfinity-stable-structures = { package = "ic-stable-structures", version = "0.6" }
-
-# version 2.1 of pocket-ic mistakenly imports native-tls which makes it incompatible with i686 testing
-# We'll switch to the official pocket-ic package once they fixed the issue
-# pocket-ic = "2.1"
-pocket-ic = { git = "https://github.com/bitfinity-network/ic", branch = "bitfinity", package = "pocket-ic" }
\ No newline at end of file
+pocket-ic = "2.2"
```
veeso, line 85: why not 3?

veeso (changes requested):

veeso (approved):

Cargo.toml
ufoscout: there's no 3

F3kilo (approved):

PR #213, opened 2024-04-11
Fix: `getrandom` dependency issue in WASM build

kobby-pentangeli:
# Motivation

The `ic-crypto-getrandom-for-wasm` package, which is a workaround for the `getrandom` dependency issues in WASM environments, doesn't work in all cases. Sometimes, just including the package in `Cargo.toml` isn't enough.

This PR provides a fix that changes how this package is consumed...
To register this new custom implementation, call the function inside of your canister's `init` method or wherever canister initialization happens, with conditional compilation, like this:

```rust
#[init]
pub fn init(&mut self) {
    #[cfg(target_family = "wasm")]
    ic_crypto_getrandom_for_wasm::register_custom_getrandom();
}
```

# Changes

- [x] Refactored implementation and export of `ic-crypto-getrandom-for-wasm::register_custom_getrandom` function.
- [x] Updated crate-level doc to reflect usage example.

# Closes

- [x] Ticket: https://infinityswap.atlassian.net/browse/EPROD-828

# NOTE

- [ ] This implementation has only been tested against a couple of canisters locally. Thorough testing against other canisters is highly recommended.

### DIFF
--- a/ic-crypto-getrandom-for-wasm/Cargo.toml
+++ b/ic-crypto-getrandom-for-wasm/Cargo.toml
@@ -4,7 +4,11 @@ version.workspace = true
 edition.workspace = true
 
 [target.'cfg(target_family = "wasm")'.dependencies]
+candid = { workspace = true }
 getrandom = { version = "0.2", features = ["custom"] }
+ic-cdk = { workspace = true }
+ic-cdk-timers = { workspace = true }
+rand = { workspace = true, features = ["getrandom"] }
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 getrandom = { version = "0.2" }
--- a/ic-crypto-getrandom-for-wasm/Cargo.toml
+++ b/ic-crypto-getrandom-for-wasm/Cargo.toml
@@ -4,12 +4,8 @@ version.workspace = true
 edition.workspace = true
 
 [target.'cfg(target_family = "wasm")'.dependencies]
-candid = { workspace = true }
 getrandom = { version = "0.2", features = ["custom"] }
-ic-cdk = { workspace = true }
-ic-cdk-timers = { workspace = true }
 rand = { workspace = true, features = ["getrandom"] }
-rand_chacha = "0.3"
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 getrandom = { version = "0.2" }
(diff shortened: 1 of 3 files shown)


veeso (approved):

ic-crypto-getrandom-for-wasm/src/lib.rs
```diff
@@ -19,20 +19,60 @@
...
+
+    use candid::Principal;
+    use rand::rngs::StdRng;
+    use rand::{RngCore, SeedableRng};
+
+    thread_local! {
+        static RNG: RefCell<Option<StdRng>> = const { RefCell::new(None) };
+    }
+
+    pub fn register_custom_getrandom() {
```
ufoscout, line 58: This is open to an entire family of possible bugs and attacks, let's discuss it carefully.

ufoscout (changes requested):

ic-crypto-getrandom-for-wasm/src/lib.rs
ufoscout: For reference: https://forum.dfinity.org/t/seeded-rng-generates-same-random-number-on-every-call/23855

kobby-pentangeli: > For reference: https://forum.dfinity.org/t/seeded-rng-generates-same-random-number-on-every-call/23855

I may be wrong here, but the concern of the OP in this thread is that he is using the rand setup to generate nonces, which are later consumed. Also, the use of thread-local storage, which ensures that each thread has its own instance of the `RNG`, provides some layer of security for maintaining isolation in a multi-threaded context.

So, it's important that we understand exactly how the `RNG` can be reproduced/predicted. @ufoscout could you please elaborate on this?

ic-crypto-getrandom-for-wasm/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition.workspace = true
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 getrandom = { version = "0.2", features = ["custom"] }
+rand = { workspace = true, features = ["getrandom"] }
```
ufoscout, line 8: to be removed

ufoscout (changes requested):

ic-crypto-getrandom-for-wasm/Cargo.toml
kobby-pentangeli: > to be removed

Yes, no longer need.
Done.

itsyaasir (approved):

F3kilo (approved):

ufoscout (approved):

PR #226, opened 2024-07-24
Fix ACL restoration after reload

Maximkaaa:
Found another bug in the log canister implementation.


ufoscout (changes requested):
Can you add a regression test for this?

Maximkaaa:
@ufoscout I did

ufoscout (approved):

PR #227, opened 2024-07-25
Fix storing the logger configuration in stable storage

Maximkaaa:
It turns out that when a canister is compiled, it cannot use two different instances of a `MemoryManager`, so the current approach to initialize the `LogState` with memory id doesn't work correctly.

This PR fixes it by getting the virtual stable memory itself from the caller instead of memory id.


ic-log/src/canister/state.rs
```diff
@@ -67,37 +42,44 @@ impl LogState {
     pub fn init(
         &mut self,
         caller: Principal,
-        memory_id: MemoryId,
+        memory: VirtualMemory<DefaultMemoryImpl>,
```
ufoscout, line 45: The memory impl should be a generic type, check the evm code where we use the same approach

ufoscout (changes requested):

ic-log/src/canister/state.rs
```diff
@@ -3,61 +3,36 @@ use std::cell::RefCell;
...
+use ic_stable_structures::{Bound, CellStructure, StableCell, Storable, VirtualMemory};
 use ic_storage::IcStorage;
 
 use crate::did::{LogCanisterError, LogCanisterSettings, LoggerAcl, LoggerPermission, Pagination};
 use crate::writer::{InMemoryWriter, Logs};
 use crate::{take_memory_records, LogSettingsV2, LoggerConfig};
 
 thread_local! {
-    static MEMORY_MANAGER: IcMemoryManager<DefaultMemoryImpl> = IcMemoryManager::init(DefaultMemoryImpl::default());
     static LOGGER_CONFIG: RefCell<Option<LoggerConfig>> = const { RefCell::new(None) };
```
ufoscout, line 14: is it not required to store the LoggerConfig too? If not, why do we need this thread local variable?

ic-log/tests/in_memory_logger.rs
ufoscout, line 1: Is the deletion of this file an error?

ic-log/src/canister/state.rs
Maximkaaa: Unfortunately, we cannot use generics in canister traits. This would break the mechanism that allows having exported function inside a dependency.

ic-log/src/canister/state.rs
Maximkaaa: `LoggerConfig` contains a reference to the filter that the `log` operations use, so that we can update the logger filter while the canister is running. We cannot (and don't need to) store it in the stable memory. We store logger filter string as a part of the `settings` field, and use it to reset `LoggerConfig` when a canister being upgraded.

ic-log/tests/in_memory_logger.rs
Maximkaaa: No, this is not an error. That test didn't actually check anything new, not covered by unit tests, but had a problem with memory initialization. I added it because I though I could use the fact that every integration test file is run in a separate process to test actual operation of the log, but since then the way the code operates changed a lot and this test is not needed any more IMO.

ufoscout (approved):

PR #231, opened 2024-09-02
[Eprod-976] Unique task IDs

ufoscout:


ic-task-scheduler/src/scheduler.rs
```diff
@@ -214,6 +221,15 @@ where
         });
     }
 
+    /// Returns the next task id.
+    fn next_task_id(&self) -> u64 {
+        let mut lock = self.task_id_sequence.lock();
+        let id = *lock.get();
+        lock.set(id + 1)
+            .expect("Unable to access the stable storage to set the next poll id");
```
veeso, line 229: task in place of poll.

Anyway, should we worry about overflowing u64 with an intense use over the time of the scheduler?

ufoscout: By processing 1 billion tasks a day, it would take 50 million years to overflow  😶

veeso (approved):

Maximkaaa (approved):
I thought we do it without changes to external API, but this way is even better (if is followed by a PR that changes the usage of the scheduler in other repos :) )

F3kilo (approved):

PR #236, opened 2024-11-16
Update dependencies - Use PocketIC 6.0

ufoscout:


F3kilo (approved):

just/code_check.just
```diff
@@ -0,0 +1,23 @@
+# Formats the code
+[group('code_check')]
+fmt args="":
+  cargo fmt --all {{args}}
```
veeso, line 4: Actually our rustfmt.toml still requires nightly, so I wouldn't keep this

veeso (changes requested):

just/code_check.just
ufoscout: This is the common fmt logic used by all of our projects. It ignores the nightly only features

veeso (approved):

PR #237, opened 2024-11-30
Fix errors when using rust-analyzer with canister method attributes

Maximkaaa:
This commit removes checks by the `query`, `update` and `init` macroses that there are duplicate methods with the same name.

The reason this causes an issue is that in contrast with the rustc, rust-analyzer expands macroses using the same process all the time. This means that lazy_static storage for the macro is not dropped between invocations which produces the error. After this change, the new declaration of the method expanded by the rust-analyzer will just replace the old one.

This will change will also make the macros' implementation more correct for 2 reasons:

1. The checks we had are not reliable. They only check for duplication inside one crate, which in most cases is already coverred by the compiler when it check that there are no 2 methods with the same name.
2. Even if there are 2 method with the same name (in the same crate or across the crates), compiler will still report an error because these methods are exported by the WASM module with the same name. Such error method is more difficult to read then the ones we had, but it is more correct and works 100% of the time, which is more important.

### DIFF
--- a/ic-auction/src/state.rs
+++ b/ic-auction/src/state.rs
@@ -211,7 +211,7 @@ pub struct Controller<'a> {
     state: &'a mut AuctionState,
 }
 
-impl<'a> Authorized<Controller<'a>> {
+impl Authorized<Controller<'_>> {
     pub fn set_min_cycles(&mut self, min_cycles: Cycles) {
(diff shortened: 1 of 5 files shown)


veeso (approved):

ufoscout (approved):
@Maximkaaa There are lint errors to be fixed

PR #239, opened 2024-12-30
[Eprod-1106] improve cache

ufoscout:


Maximkaaa (approved):

Maximkaaa:
Shall we not also add the value to the cache if there's no old value?

ic-stable-structures/src/structure/cache/btreemap.rs
```diff
@@ -51,10 +51,12 @@ where
...
     }
 
+    /// When a new value is inserted, it is also inserted into the cache; this is
+    /// required because caching on the `get` is useless in IC if the method is used in a `query` call
     fn insert(&mut self, key: K, value: V) -> Option<V> {
-        match self.inner.insert(key.clone(), value) {
+        match self.inner.insert(key.clone(), value.clone()) {
             Some(old_value) => {
-                self.cache.remove(&key);
+                self.cache.insert(key, value);
```
ufoscout, line 59: should be cached even if there was no old value in the cache

ufoscout: done

PR #240, opened 2025-01-24
EPROD-1117 pop_first and pop_last for btreemap

veeso:


ic-stable-structures/src/structure/stable_storage/btreemap.rs
```diff
@@ -46,6 +46,18 @@ where
         self.0.remove(key)
     }
 
+    fn pop_first(&mut self) -> Option<(K, V)> {
```
ufoscout, line 49: should be:
```rust
    fn pop_first(&mut self) -> Option<(K, V)> {
        self.0.pop_first()
    }
```

ic-stable-structures/src/structure/stable_storage/btreemap.rs
```diff
@@ -46,6 +46,18 @@ where
...
+        let first_key = self.first_key_value()?.0;
+
+        self.0.remove(&first_key).map(|value| (first_key, value))
+    }
+
+    fn pop_last(&mut self) -> Option<(K, V)> {
```
ufoscout, line 55: should be:
```rust
    fn pop_last(&mut self) -> Option<(K, V)> {
        self.0.pop_last()
    }
```

ufoscout (changes requested):

ufoscout (approved):

F3kilo (approved):

PR #244, opened 2025-03-07
EPROD-1159 pocket ic 8 error

veeso:
fix of errors in https://github.com/bitfinity-network/canister-sdk/pull/243


ufoscout (approved):

Maximkaaa (changes requested):
Shouldn't we also update pocket-ic dependency to 8 in this PR?

veeso:
it's already done 😅  @Maximkaaa

veeso:
it's a fix of the previous pr where the error was too generic since they changed the struct

Maximkaaa:
No, it's not: https://github.com/bitfinity-network/canister-sdk/blob/main/Cargo.toml#L86

veeso:
Ah you mean that. The latest version of the crate is 7, which is compatible with 8, because when pocket ic 8 was released, they released the 7 version

Maximkaaa (approved):

PR #245, opened 2025-03-11
EPROD-1156-live mode on pocket ic client

veeso:
Added live mode flag to be able to use the update methods for live mode instead of default

### DIFF
--- a/ic-canister-client/src/pocket_ic.rs
+++ b/ic-canister-client/src/pocket_ic.rs
@@ -13,9 +13,6 @@ pub struct PocketIcClient {
     client: Option<Arc<PocketIc>>,
     pub canister: Principal,
     pub caller: Principal,
-    /// Tells whether pocket ic instance is in live mode.
-    /// Live mode is required to use a different update call method.
-    live: bool,
 }
 
 impl PocketIcClient {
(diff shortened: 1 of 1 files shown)


F3kilo (approved):

ic-canister-client/src/pocket_ic.rs
```diff
@@ -22,6 +25,17 @@ impl PocketIcClient {
         Self::from_client(PocketIc::new().await, canister, caller)
     }
 
+    /// Creates a new instance of a PocketIcClient in live mode
+    /// The new instance is independent and have no access to canisters of other instances.
+    ///
+    /// Live mode flag is required to use a different update call method.
+    pub async fn new_live(canister: Principal, caller: Principal) -> Self {
```
ufoscout, line 32: add a method to make live an existing instance instead

ic-canister-client/src/pocket_ic.rs
```diff
@@ -32,6 +46,25 @@ impl PocketIcClient {
...
+            live: false,
+        }
+    }
+
+    /// Crates new instance of PocketIcClient from an existing client instance with live mode set.
+    ///
+    /// Note: the passed client MUST already be in live mode.
+    ///
+    /// Live mode flag is required to use a different update call method.
+    pub fn from_client_live<P: Into<Arc<PocketIc>>>(
```
ufoscout, line 58: This API can easily be misused. This should not be allowed

ic-canister-client/src/pocket_ic.rs
```diff
@@ -50,10 +83,17 @@ impl PocketIcClient {
     {
         let args = candid::encode_args(args)?;
 
-        let reply = self
-            .client()
-            .update_call(self.canister, self.caller, method, args)
-            .await?;
+        let reply = if self.live {
```
ufoscout, line 86: The `live` is not needed. From the code, it seems to me that if the client is in live mode then `client.url()` returns `Some(url)`. We can ask in the Dfinity forum whether this is the correct way to detect if a client is live.

ufoscout (changes requested):

ic-canister-client/src/pocket_ic.rs
veeso: Good catch, that is Some only in live mode

ufoscout (approved):

PR #246, opened 2025-03-25
chore: remove did generation

itsyaasir:


ufoscout (changes requested):
You moved all did generation logic to candid-extractor, but during the backlog refinement it was decided to drop the did completely

ufoscout (approved):

veeso (approved):

